PR #1, opened 2016-02-26
README: Replace tabs with spaces

wking:
For consistent indenting in the member list, fixing tabs from 50c74746
(Initial README, 2016-02-09).


caniszczyk:
Thank you!

PR #3, opened 2016-03-10
README: add a clear note about voting

philips:


philips:
On the call last week everyone suggested we needed just a quick voting guide. Here it is with references to what we can vote on.

caniszczyk:
We should also respect (6)(l) with reviews taking at least 2 weeks before a vote is called:

"Issues should be referred to the TOB sufficiently in advance of a meeting to provide an appropriate time for TOB members to evaluate the issues, and the positions of the TDC, the positions of users, as well as sufficient time to explore compromise solutions. It is expected an appropriate review should require at least a two-week review period, though it is recognized some timecritical circumstances may call for a shorter review (e.g. security issues)."

caniszczyk:
@opencontainers/tob everyone OK with this change?

wking:
On Thu, Mar 10, 2016 at 01:35:38PM -0800, Brandon Philips wrote:
>   * README: add a clear note about voting

Robert's Rules of Order suggests phrasing like “two-thirds of the
votes cast, a quorum being present” [1], although maybe the quorum bit
is too obvious to mention here.

[1]: https://archive.org/stream/Robertsrulesofor00robe_201303/robertsrulesofor00robe#page/n275/mode/2up

vbatts:
@wking http://www.theonion.com/blogpost/we-are-not-properly-following-roberts-rules-of-ord-10991

vbatts:
@caniszczyk largely OK.

@philips to be clear, to abstain means that the 2/3 vote would be of those seats that did cast a vote?

vbatts:
@philips oh it says that, and i passed right over it.

xemul:
LGTM, thanks

wking:
On Fri, Mar 11, 2016 at 07:32:08AM -0800, Vincent Batts wrote:
> @wking http://www.theonion.com/blogpost/we-are-not-properly-following-roberts-rules-of-ord-10991

Heh, fair enough.  Although I did say that it may be unnecessary [1]
because of the quorum bit in §6.m [2] (although I didn't cite the
charter in [1]).  And I think there's a useful distinction beteen
having clear rules (which I like) and pedantically enforcing those
rules (which is up to the TOB members if/when they notice an
infraction).

[1]: https://github.com/opencontainers/tob/pull/3#issuecomment-195066430
[2]: https://www.opencontainers.org/governance

PR #5, opened 2016-03-19
proposals: image-format: remove the distribution related goals

philips:
Remove these goals as they can't be done w/o distribution.


philips:
@caniszczyk you merged this in the wrong order :(

PR #9, opened 2016-04-06
Create code-of-conduct.md

RobDolinMS:
Copies the code-of-conduct file into the TOB GitHub repository so it could be included (by reference) in other OCI projects rather than being copied into (by value) each project.

Signed-off-by: Rob Dolin <RobDolin@microsoft.com>


caniszczyk:
Thanks @RobDolinMS

Issue #2, opened 2016-03-03
Update OCI charter to deal with project structure issue


caniszczyk:
See this thread:
https://groups.google.com/a/opencontainers.org/forum/#!topic/tob/1pgiEffnu0M

caniszczyk:
@opencontainers/tdc-maintainers it would be great to get any proposed solutions to this issue from the TDC. I believe the original intention was to have 1 TDC maintainer list and multiple OCI projects instead of multiple OCI projects with their respective TDC maintainers. We can move forward by updating the charter to reflect that including deciding whether each project should a) their own list of maintainers, or b) one list of maintainers across all OCI projects. We can modify the charter if need be to accommodate option a) if that is what is preferred, but b) is supported by default right now.

philips:
Agree on clarifying the governance to make option B clear (status quo)

wking:
On Thu, Mar 03, 2016 at 09:34:26AM -0800, Chris Aniszczyk wrote:
> I believe the original intention was to have 1 TDC maintainer list
> and multiple OCI projects instead of multiple OCI projects with
> their respective TDC maintainers.

That's what the draft charter looked like, but the adopted charter
moved somewhat towards multiple TDCs.  The public discussion around
that shift is in [1,2] if folks want some background reading.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Tjq5QJ_eez0
     Subject: Contribute to OCI
     Date: Mon, 3 Aug 2015 16:20:04 +0800
     Message-ID: <CAHKD89TqBQvbcdvwt-g3J--sDyq=P50qi6CytokNJznVsei7YQ@mail.gmail.com>
[2]: https://github.com/opencontainers/web/pull/4

diogomonica:
"1 TDC maintainer list and multiple OCI projects instead of multiple OCI projects with their respective TDC maintainers." seems like the right thing.

wking:
On Thu, Mar 03, 2016 at 11:04:08AM -0800, Diogo Mónica wrote:
> "1 TDC maintainer list and multiple OCI projects instead of multiple
> OCI projects with their respective TDC maintainers." seems like the
> right thing.

The initial push for multiple TDCs was so there could be a distinction
between runC maintainers and spec maintainers [1].  As long as:

* the membership, rules, and contact information of each set (TOB
  electors, TDC(s), project maintainers, …) is clear, and
* specification/testing projects are govorned independently of a
  particular implementation.

[...]

PR #15, opened 2016-06-09
proposals: add release-approval-process

philips:
This is a proposed process for approval of new releases of
specifications and projects from the OCI.

The creation of this process is designed to clarify how a release gets
created and who needs to sign off.


proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
+# OCI Project Release Approval Process v1.0
+
+OCI projects need a standard process for making releases so the community of maintainers can consistently know when something can be tagged and released. As the OCI maintains three categories of projects: specifications, applications, and conformance/testing tools we will set different rules for each.
+
+## Specifications
+
+**Planning a release:** Every OCI specification project SHOULD hold a weekly meeting that involves maintainers reviewing pull requests, debating outstanding issues, and planning releases. This meeting MUST be advertised on the project README and MAY happen on a phone call, video conference, or on IRC. Maintainers MUST send updates to the dev@opencontainers.org with results of these meetings. Maintainers MAY change the meeting cadence once a specification has reached v1.0.0. The meeting cadence MUST NOT be greater than once every four weeks The release plans, corresponding milestones and estimated due dates MUST be published on GitHub (e..g, https://github.com/opencontainers/runtime-spec/milestones).
```
wking: ”e..g” → ”e.g.”

crosbymichael: lets chill out on the nits until we look at content

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+
+**Making a release:** OCI specification projects MUST announce intentions to release with two project maintainer sponsors (listed in the repo MAINTAINERS file) on the dev@opencontainers.org mailing list. After the announcement a two-thirds super majority of project maintainers MUST reply to the list with an LGTM within one week for the release to be approved. The maintainers MUST wait a full week for maintainers to reply but if all maintainers reply with an LGTM then the release MAY release earlier except in the case of a major releases.
```
mrunalp: Maybe a different criteria for a major vs minor release will help. For e.g. 0.1 should require LGTM from half but 1.0 should require 2/3 LGTMs.

wking: The last sentence reads a bit strangely around the MUST to me.  How about replacing the last two sentences with:

> A release is approved a week after the announcement if at least two-thirds of maintainers LGTM the release.  Minor and point releases MAY be approved earlier if all maintainers LGTM the release.

And that also gives space for @mrunalp's different thresholds.

crosbymichael:
Content looks good so far but it is missing any type of quantitative criteria that tells when an release is ready or not.  It does not talk about testing phases, feedback, resolving outstanding issues, etc.

We need something that we can measure because if a maintainers that rejects a release have to state their reasons why then the maintainers whom approve must also have valid reasons to cut a release.  

i.e. "we have solved the large schema issues for the specification, no other schema changes are required and there are no outstanding bugs based on feedback from implementations, its safe for a release."

proposals/release-approval-process.md
wking: On Thu, Jun 09, 2016 at 01:48:30PM -0700, Mrunal Patel wrote:
> Maybe a different criteria for a major vs minor release will
> help. For e.g. 0.1 should require LGTM from half but 1.0 should
> require 2/3 LGTMs.

What about 1.1.0?  The charter's only Member out for “I don't like
that release” is “then you can quit” [1].  If that applies to minor
and point releases (and in the absence of wording to the contrary I
expect it does), then I think we want to be cautious about them as
well.

[1]: https://www.opencontainers.org/about/governance
     §8.d

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with a single REJECT as long two-thirds of the project maintainers approved the release.
```
wking: So if a project with seven maintainers gets 5 LGTMs and 2 REJECTs, it's rejected?  If so, I think you want to reference this rejection exception in the “Making a release” section.

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+
+**Timelines:** Specifications have a variety of different timelines in their lifecycle. In early stages the spec SHOULD release often to garner feedback. In later stages there will be bug fix releases, security fix releases, and major breaking change releases. Each of these should have a different level of notification.
+
+- Pre-v1.0.0 specifications SHOULD release on a regular cadence and MUST follow the normal release process. In practice this should be a release every week or two.
```
wking: A (bi)weekly cadence goal has never been met by an OCI project (as far as I'm aware), and it would mean you'd have to open your week-long vote for 0.2.0 immediately after cutting 0.1.0.  Can we make it “SHOULD release every month”?

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+- Major specification releases that introduce new base or optional layers, break backwards compatibility, or add non-optional features MUST release at least two release candidates spaced a minimum of one week apart. In practice this means a major release like a v1.0.0 or v2.0.0 release will take 3 weeks at minimum: one week for v1.0.0-rc1, one week for v1.0.0-rc2, and one week for v1.0.0. Maintainers SHOULD strive to make zero breaking changes during this cycle of release candidates and SHOULD add an additional release candidate when a breaking change is introduced. For example if a breaking change is introduced in -rc2 then a -rc3 SHOULD be made following the normal release process.
+- Minor releases that fix bugs, grammar, introduce optional features, tests, or tooling SHOULD be made on an as-needed basis and MUST follow the normal release process.
```
wking: The current image/runtime specs have two consumers: runtime maintainers and runtime users.  If we're semantically versioning for the users (e.g. config authors and runtime callers), then “we've added an optional field to config.json” (to use a runtime-spec example) would trigger a minor bump.  However, for runtime *maintainers*, that user-optional field is a new *requirement* that they MUST support in order to be compliant with the spec.  Since folks who are likely to maintain runtimes are also OCI Members, I suspect we want to take the same care with minor bumps that we take with major bumps.

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+- Security fix releases MUST follow a special release process that substitutes the dev@opencontainers.org email for security@opencontainers.org.
```
wking: If the concern here is leaking security holes to a public dev@ list, I'd recommend giving this the bold you're using as a quasi h3.  You don't want folks to miss it in the excitement of a security fix.

wking:
On Thu, Jun 09, 2016 at 01:52:06PM -0700, Michael Crosby wrote:
> Content looks good so far but it is missing any type of quantitative
> criteria that tells when an release is ready or not.  It does not
> talk about testing phases, feedback, resolving outstanding issues,
> etc.

Do you have to talk about that?  Isn't it up to the maintainers to
have a sufficent handle on the pulse and REJECT if they feel
concerned?

> We need something that we can measure because if a maintainers that
> rejects a release have to state their reasons why then the
> maintainers whom approve must also have valid reasons to cut a
> release.

I expect this will be covered by the release notes / ChangeLog entry.
Maybe just require that to be posted with the initial
release-intention announcement?

wking:
On Thu, Jun 09, 2016 at 02:11:07PM -0700, W. Trevor King wrote:
> I expect this will be covered by the release notes / ChangeLog
> entry.  Maybe just require that to be posted with the initial
> release-intention announcement?

This is not a big ask, because until you've compiled the release notes
/ ChangeLog entry it's hard to tell if you should be making a point,
minor, or major bump.

caniszczyk:
@crosbymichael want to add a section around release criteria then? we can iterate from there... I assume you want to include things such as at least two implementations and so on... but one way to look at it is release criteria is essentially everything defined in a milestone (e.g., https://github.com/opencontainers/runtime-spec/milestones) but if you want it to be more concrete than that, go for it! :)

crosbymichael:
@caniszczyk yep

philips:
@crosbymichael @caniszczyk I didn't put i criteria because I think they should be developed through discussions on the dev@opencontainers.org posts about releases and meetings. It should be the  responsibilities of the maintainers to define their criteria through those discussions and reflect them in issues tied to milestones.

philips:
@crosbymichael @caniszczyk if that makes sense I am happy to add some additional language into this doc.

wking:
The content and rules in this document are basically an extention of
the current maintainer's guide PR-landing rules [1] to milestone
assignment and tagging.  Should this text get moved over to a
project-template PR updating the maintainer's guide?  Should all/some
of the maintainer's move over to this repo?  If we split rules about
maintainer workflow between the two repositories, what are the rules
for what goes in each repository?

Of those choices, I prefer shifting this PR to project-template,
because that sets up the per-repo maintainer's guides that have worked
well in the past, and you don't have to give new maintainers a way to
discover this repo and its docs.  It doesn't really seem like
something that the TOB will be interested in, and the current charter
gives sole responsibility to approving releases to “the TDC” (although
that language should probably be updated to reflect whatever the
current opinions are on whether TDCs are one, many, per-project, …
[3]).

[1]: https://github.com/opencontainers/project-template/blob/1d5bddcedef071152391137716db4bdbce012767/MAINTAINERS_GUIDE.md#who-decides-what
[2]: https://www.opencontainers.org/about/governance  §5.b.vii
[3]: https://github.com/opencontainers/tob/issues/2

philips:
@wking Good idea, I am fine with putting this into project-template once we finish the discussion.

crosbymichael:
I think we have to have some type of quantitative rules or nothing has really changed.  There is no way to measure when a release should be cut right now.

If we leave it up to milestones, it is still arbitrary about what is placed or removed from a milestone.  Some of the project wide criteria can be things like, having a functional test suite to ensure the release is backwards compat with the previous release in it's major.  Having compliance tests for implementations available for the version. 

I think one of the most important things about working on a spec is stability and we need release criteria to help ensure that.

proposals/release-approval-process.md
stevvooe: There are lot of problems with restricting this to GitHub issues. What state should they be in? It calls into question how they are closed or why they were closed. The other issue is that GitHub issues can be edited and comments can be deleted. I am not sure if it is healthy to politicize the state of a GitHub issue.

Allowing the medium of rejection to affect the validity increases the chance of deflection, especially when we have multiple communication channels. A rejection should be a rejection and the reasoning can be examined afterwards.

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+
+**Timelines:** Specifications have a variety of different timelines in their lifecycle. In early stages the spec SHOULD release often to garner feedback. In later stages there will be bug fix releases, security fix releases, and major breaking change releases. Each of these should have a different level of notification.
+
+- Pre-v1.0.0 specifications SHOULD release on a regular cadence and MUST follow the normal release process. In practice this should be a release every week or two.
+- Major specification releases that introduce new base or optional layers, break backwards compatibility, or add non-optional features MUST release at least two release candidates spaced a minimum of one week apart. In practice this means a major release like a v1.0.0 or v2.0.0 release will take 3 weeks at minimum: one week for v1.0.0-rc1, one week for v1.0.0-rc2, and one week for v1.0.0. Maintainers SHOULD strive to make zero breaking changes during this cycle of release candidates and SHOULD add an additional release candidate when a breaking change is introduced. For example if a breaking change is introduced in -rc2 then a -rc3 SHOULD be made following the normal release process.
```
stevvooe: We are trying to build a specification here, not a webapp. These timelines are way too tight. Larger companies are going to take time to allocate resources and examine the specification. Time needs to be given to take appropriate community feedback.

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+- Minor releases that fix bugs, grammar, introduce optional features, tests, or tooling SHOULD be made on an as-needed basis and MUST follow the normal release process.
+- Security fix releases MUST follow a special release process that substitutes the dev@opencontainers.org email for security@opencontainers.org.
+
+## Conformance/Testing and Applications Releases
+
+**Making a release:** OCI application projects MUST announce intentions to release with two project maintainer sponsors on the dev@opencontainers.org mailing list. After the announcement at least one more project maintainer MUST reply to the list with an LGTM within two business days for the release to be approved. The maintainers SHOULD wait two business days for maintainers to reply and review. If all maintainers reply with an LGTM then the release MAY release earlier.
```
stevvooe: I hate to be silly: what is the definition of a "business day". When does it start?

proposals/release-approval-process.md
stevvooe: A release should be preceded with a release proposal, giving reasoning about how the release meets the requirements of the release. If any requirements for the release are not met, but the proposal is to proceed despite the lack of met requirements, this should be called out.

philips:
@crosbymichael I think a reason that people feel what is being placed in a milestone feels arbitrary is because they aren't following all of the individual discussions. Today most of those discussions happen on GitHub threads or on the weekly calls. That is why there is a part of this process that requires reporting out to the mailing list about the weekly progress. 

Now, I totally agree there are some base line quantitative things should be true across our projects as well, here are some of my ideas:

- Tests that cover all optional and required features of the schemas
- Tests for backwards compatibility with all post-v1.0.0 versions
- A change log outlining all of the major changes and features from the previous release
- Documentation to help an implementor through any breaking change

I think it would be helpful for people to enumerate other quantitative baseline things to make a release and I can add them to this policy. What other ideas do people have?

philips:
cc @opencontainers/image-spec-maintainers @opencontainers/runtime-spec-maintainers @opencontainers/runc-maintainers

proposals/release-approval-process.md
wking: On Thu, Jun 09, 2016 at 05:07:44PM -0700, Stephen Day wrote:
> > +**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues…
>
> There are lot of problems with restricting this to GitHub
> issues. What state should they be in? It calls into question how
> they are closed or why they were closed. The other issue is that
> GitHub issues can be edited and comments can be deleted. I am not
> sure if it is healthy to politicize the state of a GitHub issue.

I agree, but I don't think the current wording politicizes the state
(it just says concerns must be *filed* as issues).  The current
wording has the vote for tagging as the only gatekeeper, and that
makes sense to me.  The issues and milestones are just for building
consensus before a vote (possibly the next vote).

proposals/release-approval-process.md
wking: On Thu, Jun 09, 2016 at 05:26:41PM -0700, Brandon Philips wrote:
> Can you be a bit more constructive? What do you think the minimum
> time from RC to release should be assuming the community has already
> seen many earlier releases?

It didn't make it into the minutes, but I vaguely remember @RobDolinMS
floating a week and a half to turn around review at Microsoft [1].  I
think that means a “we're about to cut $MAJOR.$MINOR.$PATCH” post (as
currently suggested here), and then at least 30 days (to mirror [2]?)
for feedback/voting.  I'm fine with that 30 days being divided over 4
or 5 weekly release candidates, as long as the clock starts over if
anything more than a patch-level fix lands.  Linux's 4.6 release went
through 7 candidates before a final 4.6, with 4.6-rc1 on 2016-03-26,
4.6-rc7 on 2016-05-08, and 4.6 on 2016-05-15, so a month-ish of
candidates doesn't seem unreasonable.  With an approach like this, the
Microsoft team might only be chiming in on every other release
candidate, but they'd still have at least two rounds of review before
anything went out (which is enough time to raise alarms if anything
suspicious was in the release).

However, it is a long time to freeze development, so I suggest
starting a per-release branch as soon as the proposal goes out.  Then
master (or other releases, e.g. a bugfix on an older line) can keep
chugging along while folks decide what to do about the proposed
release.  The parallel branches would also let you freeze and propose
1.1.0-rc1 immediately after 1.0.0 lands (merging development that had
happend in another branch while 1.0.0 was under review for release).

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-06-08-17.01.log.html#l-90
[2]: https://www.opencontainers.org/about/governance
     §8.d has a 30 day limit for Members to quit the OCI after a
     release they don't like was cut.

wking:
On Thu, Jun 09, 2016 at 05:21:29PM -0700, Brandon Philips wrote:
> - Tests that cover all optional and required features of the schemas
> - Tests for backwards compatibility with all post-v1.0.0 versions

These are not going to be possible if the spec does not bundle it's
test suites (and I'm in favor of not bundling them [1,2]).

> - A change log outlining all of the major changes and features from
>   the previous release
> - Documentation to help an implementor through any breaking change

The first of these sounds critical to me [3], and the second would
certainly be nice (and I'm fine requiring it).

[1]: https://github.com/opencontainers/ocitools/issues/83
[2]: https://github.com/opencontainers/ocitools/issues/98
[3]: https://github.com/opencontainers/tob/pull/15#issuecomment-225028243

proposals/release-approval-process.md
wking: On Thu, Jun 09, 2016 at 05:13:59PM -0700, Stephen Day wrote:
> > +…at least one more project maintainer MUST reply to the list with an LGTM within two business days for the release to be approved…
> 
> I hate to be silly: what is the definition of a "business day". When
> does it start?

I don't think we want to go into that ;).  Can we just make this a
week?  Or some number of hours (96?)?

diogomonica:
I just found out about this PR from my GitHub change feed, did I miss any emails/meetings?

Feels to me that a big change like this is exactly the kind of thing that should be discussed at the TOB level, right?

proposals/release-approval-process.md
cyphar: @stevvooe Not to mention the fact that some of us are in bad timezones. :D

proposals/release-approval-process.md
cyphar: I also don't understand whether or not we're meant to announce security fixes on `dev@opencontainers.org` after they've been cut? Not to mention that security releases are generally a one-patch affair.

proposals/release-approval-process.md
cyphar: Unless "optional features" means that they are "SHOULD"s rather than "MUST"s? Like extensions to the base spec? As far as I'm aware, we don't codify extensions in the spec at the moment (and I'd hope we wouldn't start -- let's not make this spec into IRC).

proposals/release-approval-process.md
cyphar: I agree with @stevvooe, these timelines are even a bit tight for software projects. A specification bump is useless without running code, so we have to accommodate the fact that we need some cook time **with an implementation**. One week for an `rc` is alright for software, but it usually takes more than a week to review and update an implementation -- let alone test it.

wking:
On Thu, Jun 09, 2016 at 09:51:46PM -0700, Diogo Mónica wrote:
> I just found out about this PR from my GitHub change feed, did I
> miss any emails/meetings?

The triggering dev@ thread is [1], and the dev@ thread announcing this
PR is [2].

> Feels to me that a big change like this is exactly the kind of thing
> that should be discussed at the TOB level, right?

I didn't see anything in the charter about TOB oversight over releases
or the release process [3] outside of the general project
add/remove/reorg power from §6.a [4].  I'd expect the TOB to be
involved in conflict resolution (again via §6.a) if a maintainer
wanted to promote a release dispute to the TOB, but I don't see a need
for the TOB to specify release procedures (outside of conflict
resolution).  Can you explain why you feel the TOB has an interest
before a maintainer reports a conflict?

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/NPBMTrniG8Q
     Subject: Criteria for 1.0 image-spec Release
     Date: Thu, 2 Jun 2016 15:15:04 -0700 (PDT)
     Message-Id: <1f0bef4e-c786-4e64-b2b3-1634a557f4ca@opencontainers.org>
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/7OSq_UDB_tA
     Subject: Proposal: OCI Project Release Approval Process
     Date: Thu, 09 Jun 2016 20:45:20 +0000
     Message-ID: <CAD2oYtO-N2aEg+_F9BcSCZD4AG0veEFSp-6d1rqb2v0nC05WSA@mail.gmail.com>
[3]: https://github.com/opencontainers/tob/pull/15#issuecomment-225055856
[4]: https://www.opencontainers.org/about/governance

diogomonica:
@wking so what you're saying is that changes in the `opencontainers/tob` repository aren't of the concern of the TOB?

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+- Major specification releases that introduce new base or optional layers, break backwards compatibility, or add non-optional features MUST release at least two release candidates spaced a minimum of one week apart. In practice this means a major release like a v1.0.0 or v2.0.0 release will take 3 weeks at minimum: one week for v1.0.0-rc1, one week for v1.0.0-rc2, and one week for v1.0.0. Maintainers SHOULD strive to make zero breaking changes during this cycle of release candidates and SHOULD add an additional release candidate when a breaking change is introduced. For example if a breaking change is introduced in -rc2 then a -rc3 SHOULD be made following the normal release process.
+- Minor releases that fix bugs, grammar, introduce optional features, tests, or tooling SHOULD be made on an as-needed basis and MUST follow the normal release process.
+- Security fix releases MUST follow a special release process that substitutes the dev@opencontainers.org email for security@opencontainers.org.
+
+## Conformance/Testing and Applications Releases
```
cyphar: This section is much shorter than the specification section. We should definitely codify rc releases -- this is quite important for several reasons. Currently we have runc `1.0.0-rc1` but I feel like there's quite a few things left to do and I'm not sure how many rcs people plan to have. I do like the fact that alerts on dev@opencontainers.org is required (I actually didn't know about the `rc1` bump until I did `git pull`).

proposals/release-approval-process.md
wking: On Thu, Jun 09, 2016 at 10:02:07PM -0700, Aleksa Sarai wrote:
> Unless "optional features" means that they are "SHOULD"s rather than
> "MUST"s? Like extensions to the base spec? As far as I'm aware, we
> don't codify extensions in the spec at the moment (and I'd hope we
> wouldn't start -- let's not make this spec into IRC).

The runtime-spec has seven SHOULDs in config.md and config-linux.md as
of v1.0.0-rc1, and README language tying those to an “unconditionally
compliant” label [1].  So I think we want to give folks as much time
to chime in on a new SHOULD as we would for a new MUST.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/README.md#open-container-runtime-specification

wking:
On Thu, Jun 09, 2016 at 10:08:16PM -0700, Diogo Mónica wrote:
> @wking so what you're saying is that changes in the
> `opencontainers/tob` repository aren't of the concern of the TOB?

They absolutely are.  I'm just saying this PR was misdirected [1], and
@philips agrees it should be moved to project-tempate after discussion
settles [2].  If the discussion is to spammy for TOB members, I'm fine
moving the PR before discussion settles (we can always link back).

[1]: https://github.com/opencontainers/tob/pull/15#issuecomment-225055856
[2]: https://github.com/opencontainers/tob/pull/15#issuecomment-225056407

proposals/release-approval-process.md
cyphar: It also should be a MUST, and it should definitely be a week.

cyphar:
@wking Yeah, I was also confused why it wasn't in project-template -- that's where we codify the rules of maintenance and contribution. Though, to be fair, TOB should be aware of what rules we're putting in place.

proposals/release-approval-process.md
wking: On Thu, Jun 09, 2016 at 10:11:28PM -0700, Aleksa Sarai wrote:
> This section is much shorter than the specification section. We
> should definitely codify rc releases -- this is quite important for
> several reasons.

For example…?  The reason why it's shorter/simpler is that OCI Members
aren't locked into supporting runC, while they are locked into
supporting OCI specs [1].  I'm not clear what sort of obligations a
runtime-spec release would impose on a Member, but I don't see how a
runC release would impose any obligations.

[1]: https://www.opencontainers.org/about/governance

cyphar:
@philips 

> Today most of those discussions happen on GitHub threads or on the weekly calls. That is why there is a part of this process that requires reporting out to the mailing list about the weekly progress.

While I do follow almost all GitHub threads, I cannot attend the weekly calls at the moment (they are at **3 am** in my timezone). It would be great if we could reschedule the weekly calls so that people in unfortunate timezones (like me) can contribute. Maybe even do a round-robin where the weekly calls are sometimes inconvenient for me and sometimes inconvenient for others.

wking:
On Thu, Jun 09, 2016 at 10:20:32PM -0700, Aleksa Sarai wrote:
> … It would be great if we could reschedule the weekly calls so that
> people in unfortunate timezones (like me) can contribute…

There's already a lot going on in this PR, and the weekly calls aren't
terribly important in this proposal (which mainly has teeth in email
voting).  Can we punt the call rescheduling to [1]?

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/HXO-FplpEKM
     Subject: OCI Call Time
     Date: Wed, 1 Jun 2016 21:10:21 +1000
     Message-ID: <591bb0a3-278e-f82c-6514-ff6a35d8d579@suse.de>

proposals/release-approval-process.md
cyphar: I'm confused -- why is this in here at all then? runC is an OCI project, so we should be also making the rules clear for runC as well as the *-spec projects. If we don't want to make any rules for runC then we should remove the "Applications" from the title. No rules are better than half-baked rules.

wking: On Thu, Jun 09, 2016 at 10:30:22PM -0700, Aleksa Sarai wrote:
> > +## Conformance/Testing and Applications Releases
>
> I'm confused -- why is this in here at all then? … No rules are
> better than half-baked rules.

I'm all in favor of fully baking the rules.  I'm just trying to
explain why they may not need to be as strict as the spec rules.  I
agree that runC rcs would be nice if cutting a runC release meant
committing to bug-level backwards compat.  But if the compatibility is
covered on the one side by runtime-spec and the other by a
command-line API spec (or similar), then the only remaining consumer
exposure is out-of-spec runC extensions.  If those get the same OCI
weight as in-spec features, then yeah, runC extensions are basically
another spec repository.  If those get less weight than in-spec
features, then you can probably cut a few corners.

proposals/release-approval-process.md
RobDolinMS: @wking Is right here.  The runtime-spec has hit about monthly milestones.  I'm a +1 making monthly the minimum.

proposals/release-approval-process.md
philips: Are the @opencontainers/runc-maintainers OK with a week?

proposals/release-approval-process.md
vishh: It will be useful to identify a few key stakeholders beyond just maintainers to validate a release candidate. Just having one or two implementations doesn't seem to satisfy the requirements of a standard. Would it be possible to come up with a list of partners who have a stake in this effort and get their feedback before making a final release?

If we do end up identifying some partners, then we will have to give them sufficient time to validate a release candidate. For large organizations, I would not be surprised if this process takes a month.

proposals/release-approval-process.md
anuthan: I would also request that we get a LGTM from one of the platform specific parties for "Making a release". As in atleast one LGTM for Linux , one from Windows and one from Solaris or whoever, is contributing to the specified project. We can keep a time limit for this response as well.

proposals/release-approval-process.md
wking: On Fri, Jun 10, 2016 at 12:12:29PM -0700, Vish Kannan wrote:
> Would it be possible to come up with a list of partners who have a
> stake in this effort and get their feedback before making a final
> release?

Isn't there already a list of pretty icons for that [1]?  I think the
goal is to make it clear where folks who have an interest should watch
for announcements, and to tell them how to chime in with feedback, but
I don't think we need to go around and knock on doors.

> If we do end up identifying some partners, then we will have to give
> them sufficient time to validate a release candidate. For large
> organizations, I would not be surprised if this process takes a
> month.

A month (4 weekly rcs + a final release) works for me [2], but in that
case you'd expect the slow responders to be showing up at the last
minute with any feedback.  Still, the charter only gives Members 30
days to quit after a release [2], so I think it's safe to assume
everone involved can turn things around in that time frame.

[1]: https://www.opencontainers.org/about/members
[2]: https://github.com/opencontainers/tob/pull/15#discussion_r66559230

proposals/release-approval-process.md
wking: On Fri, Jun 10, 2016 at 12:17:22PM -0700, Abhijeeth Nuthan wrote:
> I would also request that we get a LGTM from one of the platform
> specific parties for "Making a release". As in atleast one LGTM for
> Linux , one from Windows and one from Solaris or whoever, is
> contributing to the specified project.

I don't think there's a need to call that out.  Platforms who wish to
be represented should have maintainers they trust (and/or employ) on
the project, and expect the maintainers to respect any per-platform
concerns raised (just as concerns from anybody should be respected,
#17).

proposals/release-approval-process.md
```diff
@@ -0,0 +1,26 @@
...
+
+**Making a release:** OCI application projects MUST announce intentions to release with two project maintainer sponsors on the dev@opencontainers.org mailing list. After the announcement at least one more project maintainer MUST reply to the list with an LGTM within two business days for the release to be approved. The maintainers SHOULD wait two business days for maintainers to reply and review. If all maintainers reply with an LGTM then the release MAY release earlier.
+
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with a single REJECT.
```
diogomonica: What happens when there are two REJECTS? Do we stay in the "resolve the concerns and wait" reconciliation loop? Is there a timeout where the issue is then raised to the TOB?

wking: On Fri, Jun 10, 2016 at 12:25:11PM -0700, Diogo Mónica wrote:
> What happens when there are two REJECTS?

More on this here [1].

> Do we stay in the "resolve the concerns and wait" reconciliation
> loop? Is there a timeout where the issue is then raised to the TOB?

I think this should be “at their discretion, any maintainer may appeal
a disagreement to the TOB for mediation”, which is probably already
covered by the charter's §6.a [2].

[1]: https://github.com/opencontainers/tob/pull/15/files/bdfa70d70f093146bc730be2576586ec8ed57cca#r66519789
[2]: https://www.opencontainers.org/about/governance

proposals/release-approval-process.md
stevvooe: @philips A 1 month minimum sounds okay. I would still like to hear from the larger companies on this.

crosbymichael:
FYI: I'm still working on some proposals for adding additional criteria for our release process.  Just a little busy with my day job this week but I'll try to PR something within the next few days.

proposals/release-approval-process.md
philips: I think the members of security@opencontainers.org team can figure out and coordinate the disclosure process to dev@. If someone wants to document this security release process we can handle it in another document as I don't think it is in the critical path to clarifying our overall release process here.

proposals/release-approval-process.md
philips: I think the best course here so things are easy to understand is that we say 2/3 LGTM vote for a release is sufficient. But, rejects should be seriously considered and time and discussion made to change the REJECT to an LGTM. I will fix this up in a follow-up commit.

philips:
Thanks for all of the feedback everyone, I am going to put up some new commits that make the following changes:

- Requiring applications wait 1 week to get feedback before making a release, removing "business day" wording @cyphar, @stevvooe, and @wking were [in that discussion](https://github.com/opencontainers/tob/pull/15#discussion_r66543927).
- Requiring the _minimum_ process for a major release to be 3 rcs and a final release. This will establish a _minimum_ timeline of 1 month to get to a release assuming zero required changes. A breaking change would "reset the clock" on the rc process @stevvooe, @wking were [in that discussion](https://github.com/opencontainers/tob/pull/15#discussion_r66543579)
- Changing the release goal for projects to a "SHOULD monthly release" from the original bi-weekly. @diogomonica, @stevvooe, @mrunalp, @RobDolinMS were [in that discussion](https://github.com/opencontainers/tob/pull/15#discussion_r66520187)
- Fix up the language around REJECTs so it is easier to understand. The basic premise is that a release may continue with REJECTs if 2/3 of the maintainers vote to make the release. But, the maintainers SHOULD discuss and allow time for any REJECTs to become LGTMs. Spread over two discussions: [1](https://github.com/opencontainers/tob/pull/15/files/bdfa70d70f093146bc730be2576586ec8ed57cca#r66519789) and [2](https://github.com/opencontainers/tob/pull/15/files/bdfa70d70f093146bc730be2576586ec8ed57cca#r66668148)

If you feel I am missing a fixup please let me know. I will try and get these changes made in a few hours for review. Each will be a separate commit.

philips:
Michael- do you mind if I try to put a release criteria proposal together
based on your ideas and mine today? I would like to have something to
discuss on the Wednesday call.

wking:
On Tue, Jun 14, 2016 at 08:20:40AM -0700, Brandon Philips wrote:
> If you feel I am missing a fixup please let me know.

* Major vs. minor vs. patch releases [1] and optional features [2]
  (although adjusting this document should probably wait for clarity
  in #16).
* Requiring ChangeLog entries and/or release notes with the release
  proposal [3,4], although this may be part of your planned release
  criteria [5].
* Explicitly saying that the state of GitHub milestones and issues is
  informative, and that the “do we cut $COMMIT as $VERSION” call
  depends solely on maintainer votes (avoiding confusion like [6]).
* Deciding whether non-maintainers should be actively consulted
  (vs. responding to non-maintainer-initiated feedback) [7].  This is
  probably covered by #17.
* Deciding whether per-platform LGTMs are required [8].  I don't think
  we need this [9], but nobody else has chimed in on the issue.

[1]: https://github.com/opencontainers/tob/pull/15/files/424173b865452ec4c421a784988754202f486cf4#r66518877
[2]: https://github.com/opencontainers/tob/pull/15/files/bdfa70d70f093146bc730be2576586ec8ed57cca#r66562016
[3]: https://github.com/opencontainers/tob/pull/15#issuecomment-225027915
[4]: https://github.com/opencontainers/tob/pull/15/files/424173b865452ec4c421a784988754202f486cf4#r66544287
[5]: https://github.com/opencontainers/tob/pull/15#issuecomment-225918589
[6]: https://github.com/opencontainers/tob/pull/15/files/bdfa70d70f093146bc730be2576586ec8ed57cca#r66543381
[7]: https://github.com/opencontainers/tob/pull/15/files/bdfa70d70f093146bc730be2576586ec8ed57cca#r66666520
[8]: https://github.com/opencontainers/tob/pull/15/files/424173b865452ec4c421a784988754202f486cf4#r66667191
[9]: https://github.com/opencontainers/tob/pull/15#discussion_r66667645

proposals/release-approval-process.md
```diff
@@ -10,7 +10,7 @@ This approval process hopes to encourage early consistent consensus building dur
...
 
-**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with a single REJECT as long two-thirds of the project maintainers approved the release.
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long two-thirds of the project maintainers approved the release.
```
wking: Just to make sure, you do intend this to be “⅔ of maintainers” and not “⅔ of voting maintainers”?  For example, a seven-maintainer project that gets 4 LGTMs and no REJECTs should not be released.  For comparison, TOB votes are ⅔ of votes cast (#3).

philips: @wking good point, it should be voting maintainers. We have nearly hit this corner case with Vincent, Brendan, and Jon all being on vacation at the same time a few weeks ago.

proposals/release-approval-process.md
philips: I agree we don't need to call this out. Maintainers should be trusted to make the right call, and platform maintainers should make their case to the relevant channels. It is pretty clear how this stuff should work based on the OCI gov. docs 5.d: "The primary means for any organization to influence the technical direction of the OCI Projects is via contribution or service as maintainers. "

proposals/release-approval-process.md
wking: On Tue, Jun 14, 2016 at 12:14:01PM -0700, Brandon Philips wrote:
> @wking good point, it should be voting maintainers. We have nearly
> hit this corner case with Vincent, Brendan, and Jon all being on
> vacation at the same time a few weeks ago.

In that case, you probably also want language establishing a quorum,
to avoid rogue rump releases ;).

philips: Fixed in d6a6dbe

philips:
If you have been following along you can see a number of pieces of feedback addressed in commits made today, June 14th: https://github.com/opencontainers/tob/pull/15/commits

philips:
On Tue, Jun 14, 2016 at 11:48 AM W. Trevor King <notifications@github.com>
wrote:

> On Tue, Jun 14, 2016 at 08:20:40AM -0700, Brandon Philips wrote:
> > If you feel I am missing a fixup please let me know.
>
> * Major vs. minor vs. patch releases [1] and optional features [2]
> (although adjusting this document should probably wait for clarity
> in #16).
>

Yes.


> * Requiring ChangeLog entries and/or release notes with the release
> proposal [3,4], although this may be part of your planned release
> criteria [5].
>

Yes, this should be part of the criteria guidelines.


> * Explicitly saying that the state of GitHub milestones and issues is
> informative, and that the “do we cut $COMMIT as $VERSION” call
> depends solely on maintainer votes (avoiding confusion like [6]).
>

Fixed in 86a3255


> * Deciding whether non-maintainers should be actively consulted
> (vs. responding to non-maintainer-initiated feedback) [7]. This is
> probably covered by #17.
>

Yes, we should put this into contributing.md template


> * Deciding whether per-platform LGTMs are required [8]. I don't think
> we need this [9], but nobody else has chimed in on the issue.
>

Commented with my opinion.

>

proposals/release-approval-process.md
```diff
@@ -8,9 +8,9 @@ This approval process hopes to encourage early consistent consensus building dur
...
 
-**Making a release:** OCI specification projects MUST announce intentions to release with two project maintainer sponsors (listed in the repo MAINTAINERS file) on the dev@opencontainers.org mailing list. After the announcement a two-thirds super majority of project maintainers MUST reply to the list with an LGTM within one week for the release to be approved. The maintainers MUST wait a full week for maintainers to reply but if all maintainers reply with an LGTM then the release MAY release earlier except in the case of a major releases.
+**Making a release:** OCI specification projects MUST announce intentions to release with two project maintainer sponsors (listed in the repo MAINTAINERS file) on the dev@opencontainers.org mailing list. After the announcement two-thirds of the voting project maintainers MUST reply to the list with an LGTM within one week for the release to be approved. The maintainers MUST wait a full week for maintainers to reply but if all maintainers reply with an LGTM then the release MAY release earlier except in the case of a major releases.
```
wking: This still [doesn't have quorum language][1].  And [I do like Robert's “two-thirds of the votes cast, a quorum being present”][2].  How about:

> Voting on proposed releases should happen on the appropriate mailing list with maintainers posting LGTM or REJECT.  They may also explicitly not vote by posting ABSTAIN (which is useful to revert a previous vote).  Maintainers may post multiple times (e.g. as they revise their position based on feeback), but only their final post counts as their vote.  A proposed release passes if two-thirds of votes cast, a quorum having voted, are in favor of the release.  A quorum is established when at two-thirds of maintainers have voted.  Voting SHOULD remain open for a week, although under exceptional conditions (e.g. security fixes) non-major releases which reach quorum with unanimous support MAY be released earlier.

[1]: https://github.com/opencontainers/tob/pull/15#discussion_r67037985
[2]:https://github.com/opencontainers/tob/pull/3#issuecomment-195066430

philips: ack, I see what you are saying now. I will add some more robust two-thirds language based on your suggestion later this afternoon.

proposals/release-approval-process.md
stevvooe: Are we continuing to double-down on only using github issues?

My experience is that they are ripe for deflection. The only requirement is that they should be in written form. If that is too open, at least allow the mailing list as a forum.

Or is your goal to prevent rejections happening over the mailing list in the future?

wking: On Tue, Jun 14, 2016 at 01:10:32PM -0700, Stephen Day wrote:
> Are we continuing to double-down on only using github issues?

I think the role of GitHub issues should be discussed on 86a32553.

wking: On Tue, Jun 14, 2016 at 01:13:52PM -0700, W. Trevor King wrote:
> Tue, Jun 14, 2016 at 01:10:32PM -0700, Stephen Day:
> > Are we continuing to double-down on only using github issues?
> 
> I think the role of GitHub issues should be discussed on 86a32553.

Ah nevermind, there's probably not a good place to comment on the
rejection-reason requirement for GitHub issues.  I'm fine with
allowing the mailing list as well.

proposals/release-approval-process.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Making a release:** OCI specification projects MUST announce intentions to release with two project maintainer sponsors (listed in the repo MAINTAINERS file) on the dev@opencontainers.org mailing list. After the announcement two-thirds of the voting project maintainers MUST reply to the list with an LGTM within one week for the release to be approved. The maintainers MUST wait a full week for maintainers to reply but if all maintainers reply with an LGTM then the release MAY release earlier except in the case of a major releases.
+
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long two-thirds of the voting project maintainers approved the release.
```
mrunalp: as long as..

mrunalp:
Overall this looks fine.

dqminh:
Do we want to reformat to follow the 1 line per sentence pattern in the spec ?

philips:
@dqminh Good point, I will do a reformat once everyone is happy with the state.

Added the final two changes, I think, to this branch:

- Establish quorum rules: 33d5a19c59aa1bd6da9fb25b64fee1d658a73cc3
- Allow maintainers to include concerns for a release in GitHub or inline email: 78d6c1ec6ba0e0285ff1b61276e05dbafa427d3b

Based on the Weekly Call today this should be the last remaining feedback on the OCI release process assuming these changes are acceptable.

Please review with feedback or an LGTM @opencontainers/image-spec-maintainers and @opencontainers/runtime-spec-maintainers and @opencontainers/runc-maintainers! We want to get this in place so we can start trying the process out and making regular releases.

Also, on today's call @crosbymichael and I also discussed adding release criteria but are going to make that a separate document or followup so we can give this new release process some time to be beta tested before concerning ourselves with the criteria for a v1.0.0 release of the various specs. For those who have not been following that discussion it will roughly look like this: https://github.com/opencontainers/tob/pull/15#issuecomment-225063969

proposals/release-approval-process.md
```diff
@@ -8,9 +8,9 @@ This approval process hopes to encourage early consistent consensus building dur
...
+**Making a release:** OCI specification projects MUST announce intentions to release with two project maintainer sponsors (listed in the repo MAINTAINERS file) on the dev@opencontainers.org mailing list. Voting on proposed releases should happen on the dev@opencontainers.org mailing list with maintainers posting LGTM or REJECT. Maintainers may also explicitly not vote by posting ABSTAIN (which is useful to revert a previous vote). Maintainers may post multiple times (e.g. as they revise their position based on feeback), but only their final post counts as their vote. A proposed release passes if two-thirds of votes cast, a quorum having voted, are in favor of the release. A quorum is established when at two-thirds of maintainers have voted. Voting SHOULD remain open for a week, although under exceptional conditions (e.g. security fixes) non-major releases which reach quorum with unanimous support MAY be released earlier.
```
wking: You probably want quorum language for the “Conformance/Testing and Applications Releases” section as well.  It might be easier to collect the shared stuff in a separate section, and then have sub-sections to call out the spec-specific and app-specific adjustments.

philips: @wking Hrm, oops, I meant to add that to the apps section.

proposals/release-approval-process.md
```diff
@@ -8,9 +8,9 @@ This approval process hopes to encourage early consistent consensus building dur
...
 
-**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long two-thirds of the voting project maintainers approved the release.
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long as a quorom has been established and two-thirds of the voting project maintainers approved the release. A quorum is established when at two-thirds of maintainers have voted. Voting SHOULD remain open for a week, although under exceptional conditions (e.g. security fixes) non-major releases which reach quorum with unanimous support MAY be released earlier.
```
wking: The previous paragraph discusses how REJECT votes impact vote success (and in fact, a REJECT vote may help *pass* a release by pushing it over quorum).  I think this paragraph should restrict itself to “how should discussion move forward to account for REJECT votes”, e.g. rejecter lists concerns, discussion ensues, other maintainers may choose to adjust their earlier vote (as long as the voting is still open).  Repeating voting rules in this paragraph doesn't seem useful.

proposals/release-approval-process.md
```diff
@@ -10,7 +10,7 @@ This approval process hopes to encourage early consistent consensus building dur
...
 
-**Rejecting a release:** A project maintainer MAY choose to reply with REJECT and MUST include a list of concerns filed as GitHub issues or inline in the mailing list reply. The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long as a quorom has been established and two-thirds of the voting project maintainers approved the release. A quorum is established when at two-thirds of maintainers have voted.
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT. A project maintainer posting a REJECT MUST include a list of concerns or links to written documentation for those concerns (e.g. GitHub issues or mailing-list threads). The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long as a quorom has been established and two-thirds of the voting project maintainers approved the release. A quorum is established when at two-thirds of maintainers have voted.
```
wking: “when at two-thirds” → “when at least two-thirds” (and you have this typo below as well).

I [still][1] think we don't need to reiterate the quorum requirements and other voting nits in this paragraph, since they were covered in the previous paragraph.

[1]: https://github.com/opencontainers/tob/pull/15#discussion_r67241646

proposals/release-approval-process.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Rejecting a release:** A project maintainer MAY choose to reply with REJECT. A project maintainer posting a REJECT MUST include a list of concerns or links to written documentation for those concerns (e.g. GitHub issues or mailing-list threads). The project maintainers SHOULD try to resolve the concerns and wait for the rejecting maintainer to change their opinion to LGTM. However, a release MAY continue with REJECTs as long as a quorom has been established and two-thirds of the voting project maintainers approved the release. A quorum is established when at least two-thirds of maintainers have voted. Voting SHOULD remain open for a week, although under exceptional conditions (e.g. security fixes) non-major releases which reach quorum with unanimous support MAY be released earlier.
```
wking: In the specs section, the quorum rules are part of “Making a release”, but here you have them under “Rejecting a release”.  I [still][1] think we should collect the shared “how do we vote on the mailing list”  rules in a single section, so we don't have to worry about consistency between these two copies.

[1]: https://github.com/opencontainers/tob/pull/15#discussion_r67241092

wking: For example, as of fb003ff4, only spec maintainers can ABSTAIN.  If that's a useful concept for spec maintainers, I expect testing/app maintainers will want it too.

philips: It is hard to make shared rules as the consensus so far is that apps need a
"lighter weight" process than specs. So the app process has a lower bar for
a release but keeps the same bar for resolution of REJECTs.

Open to suggestions on how to make it less awkward but I think it is OK.

wking: On Wed, Jun 15, 2016 at 09:28:30PM -0700, Brandon Philips wrote:
> It is hard to make shared rules as the consensus so far is that apps
> need a "lighter weight" process than specs. So the app process has a
> lower bar for a release but keeps the same bar for resolution of
> REJECTs.

I've filed philips/tob#1 with a stab at a DRY structure.  That gives
you all of your current list-based voting rules for all projects, with
the spec-specific rc rules and timeline requirements that you have
here.  I didn't see any other app/spec differences in your version,
but let me know if I missed any.

hqhq:
It looks good to me.

proposals/release-approval-process.md
```diff
@@ -0,0 +1,38 @@
+# OCI Project Release Approval Process v1.0
+
+OCI projects need a standard process for making releases so the community of maintainers can consistently know when something can be tagged and released. This approval process hopes to encourage early consistent consensus building during project and specification development. The mechanisms used are regular community communication on the mailing list about progress, scheduled meetings for issue resolution and release triage, and regularly paced and communicated releases. An anti-pattern that we want to avoid is heavy development or discussions "late cycle" around major releases. We want do build a community that is involved and communicates consistently through all releases instead of relying on "silent periods" as a judge of stability.
```
mrunalp: s/We want do/We want to/

proposals/release-approval-process.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Timelines:** Specifications have a variety of different timelines in their lifecycle.
+
+- Pre-v1.0.0 specifications SHOULD release on a monthly cadence to garner feedback.
+- Major specification releases MUST release at least three release candidates spaced a minimum of one week apart. This means a major release like a v1.0.0 or v2.0.0 release will take 1 month at minimum: one week for rc1, one week for rc2, one week for rc3, and one week for the major release itself. Maintainers SHOULD strive to make zero breaking changes during this cycle of release candidates and SHOULD add restart the three-candidate count when a breaking change is introduced. For example if a breaking change is introduced in v1.0.0-rc2 then the series would end with v1.0.0-rc4 and v1.0.0.
```
tianon: Minor typo? "... release candidates and SHOULD **add** restart the three-candidate count when a breaking ..."

proposals/release-approval-process.md
philips: done.

proposals/release-approval-process.md
philips: fixed

wking:
I think the best way to address Jason's (paraphrased) “[how do we
amend this?][1]” is by separating project governance (with generic
voting rules and such) from release-specific rules.  I've taken a stab
at that in philips/tob#2, if folks want to look it over.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!msg/dev/ik3MIDWq4Us/Zx1JUStXBAAJ

proposals/release-approval-process.md
```diff
@@ -10,7 +10,7 @@ OCI projects need a standard process for making releases so the community of mai
 
 ## Security fixes
 
-Security fix releases MUST use security@opencontainers.org instead of dev@opencontainers.org, but should otherwise follow the standard [list-based voting process](#list-based-voting).
+Security fix releases MUST use security@opencontainers.org instead of dev@opencontainers.org, but should otherwise follow the standard [list-based voting process](#list-based-voting). The security@opencontainers.org email includes all members of the TOB; the TOB will guide the security sensitive release with project maintainers.
```
wking: I tried to flesh this out a bit in philips/tob#2, but I'm concerned about two things:

* If you pull in the TOB, you have to explain how the TOB and maintainers will interact (i.e. what does “guide” mean in your sentence?).  I'd rather just have {project}+security@ addresses (e.g. runtime-spec+security@opencontainers.org), and the maintainers can pull in the TOB as they see fit (like always).

* Who gets advanced notification of security fixes?  Do folks who want advanced access apply by mailing the security address?  E.g. “please let us know before pushing security fixes so we can patch our system.  We have $N users, and you can trust us not to leak/abuse the information because $REASONS”.  Where is the list of approved early-contacts maintained (probably not in the public project repo)?

proposals/release-approval-process.md
```diff
@@ -0,0 +1,38 @@
...
+* They are subject to the OFWa patent grant (§8.d and e).
+
+To avoid unfortunate side effects (onerous backwards compatibity requirements or Member resignations), the following additional procedures apply to specification releases:
+
+**Planning a release:** Every OCI specification project SHOULD hold meetings that involves maintainers reviewing pull requests, debating outstanding issues, and planning releases. This meeting MUST be advertised on the project README and MAY happen on a phone call, video conference, or on IRC. Maintainers MUST send updates to the dev@opencontainers.org with results of these meetings. Before the specification reaches v1.0.0, the meetings SHOULD be weekly.  Once a specification has reached v1.0.0, the maintainers may alter the cadence, but the meeting cadence MUST NOT be greater than once every four weeks. The release plans, corresponding milestones and estimated due dates MUST be published on GitHub (e.g. https://github.com/opencontainers/runtime-spec/milestones). GitHub milestones and issues are only used for community organization and all releases MUST follow the [list-based voting process](#list-based-voting).
```
vbatts: Is this saying that immediately after releasing v1.0.0 that the weekly call will go to once a month?

wking: On Mon, Jun 27, 2016 at 05:44:09PM -0700, Vincent Batts wrote:
> > +**Planning a release:** Every OCI specification project SHOULD
> >  hold meetings that involves maintainers reviewing pull requests,
> >  debating outstanding issues, and planning releases… Before the
> >  specification reaches v1.0.0, the meetings SHOULD be weekly.
> >  Once a specification has reached v1.0.0, the maintainers may
> >  alter the cadence, but the meeting cadence MUST NOT be greater
> >  than once every four weeks…
>
> Is this saying that immediately after releasing v1.0.0 that the
> weekly call will go to once a month?

What gives you that impression?  I read it as:

a. Maintainers don't have to hold meetings ever if they don't want
   (initial SHOULD in my excerpt).
b. Before 1.0, this document recommends spec-project meetings weekly.
c. After 1.0, this document does not have a recommended meeting
   frequency.
d. After 1.0, this document sets a maximum meeting frequency of 28
   days.

So after 1.0, the project maintainers may opt to transition to four
weeks, but they don't have to.

I agree that the section could be reworded for increased clarity
though.  Having a MUST limit on frequency (that only applies after
1.0?) is strange if having meetings at all is only a SHOULD.  I'll
file a PR against @philips branch (if he doesn't update it on his own)
once I see how philips/tob#2 works out.

cyphar:
Something that isn't mentioned is signed tags of releases. This is something we really need to get a handle on, with each maintainer having a GPG key and they must sign releases. I personally have shifted to signing each individual commit, but that's optional.

proposals/release-approval-process.md
RobDolinMS: >the meeting cadence MUST NOT be greater than once every four weeks
I think more clear language might be:
the gap between meetings MUST NOT be greater than four weeks
/cc @Philips, @vbatts, @wking

wking: On Wed, Jun 29, 2016 at 11:30:45AM -0700, Rob Dolin (MSFT) wrote:
> >the meeting cadence MUST NOT be greater than once every four weeks
>
> I think more clear language might be:
> the gap between meetings MUST NOT be greater than four weeks

I think the tip of this discussion has shifted to [1] (no PR for that
yet?) [2].  The current wording there tried to address this issue [3],
but I'm fine with the wording you're proposing too.

[1]: https://github.com/philips/project-template/commits/add-governance-and-releases-docs
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/ik3MIDWq4Us/4OPdr-iWBQAJ
     Subject: Re: Vote Required: OCI Image Spec Release Process
     Date: Wed, 29 Jun 2016 01:38:07 +0000
     Message-ID: <CAD2oYtPazRCBNHRHGyUt=wAA+=fber0AErqz5rnf4irc=QX9pQ@mail.gmail.com>
[3]: https://github.com/philips/project-template/commit/c9a01f5fb1c59c91d7ebba624f0dc6970f863d6c#diff-5cafcf46bfb947f8e58fa5c16d39018cR34

caniszczyk:
@philips what's left on this to finalize this?

philips:
@caniszczyk As I think we are getting close to having a policy [everyone agrees with](https://groups.google.com/a/opencontainers.org/d/msg/dev/ik3MIDWq4Us/GhsS6MPiBwAJ). So, I just put up a new PR against the [project-template repo](https://github.com/opencontainers/project-template/pull/15) and will post to the mailing list for a new vote.

philips:
Vote has been sent to project maintainers: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/x-Oh3PDz1Y8 and the TOB has been cc'd. Closing this PR out.

PR #26, opened 2017-02-06
proposals: Update links to the scope table

wking:
At some point this changed from `/governance/` to `/about/`.  @caniszczyk just [updated the charter link][1], so the `/about/` version is probably canonical.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-06.log.html#t2017-02-06T17:48:07


caniszczyk:
Thanks!

PR #28, opened 2017-02-09
fix style of markdown files

AkihiroSuda:
Before:
before

After:
after

Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>


caniszczyk:
Thanks!

PR #22, opened 2016-12-20
Add OCI digest project proposal

caniszczyk:
For record keeping, add project proposal for go-digest:
https://github.com/opencontainers/tob/issues/21

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

PR #32, opened 2017-08-04
Add proposal for libocispec

giuseppe:
Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ https://groups.google.com/a/opencontainers.org/forum/#!forum/tob (tob@opencontai
 * [Digest](proposals/digest.md)
 * [Distribution Spec](proposals/distribution.md)
 * [Image Format Spec](proposals/image-format)
+* [libocispec](proposals/libocispec.md)
 * [SELinux](proposals/selinux.md)
 * [Tools](proposals/tools.md)
 
--- a/proposals/libocispec.md
+++ b/proposals/libocispec.md
@@ -1,0 +1,36 @@
+# OCI libocispec project proposal
+
+## Abstract
+Provide a C library for parsing and accessing OCI Runtime and Image specification JSON.
+
+## Proposal
+With repositories under the https://github.com/opencontainers organization:
+
+Rename (transfer) https://github.com/giuseppe/libocispec to https://github.com/opencontainers/libocispec.
+
+### Initial Maintainers
+Initial maintainers of the libocispec project:
+* Giuseppe Scrivano <gscrivan@redhat.com> (@giuseppe)
+* Yifeng Tan <tanyifeng1@huawei.com> (@tanyifeng)
+
+### Code of Conduct
+This project would incorporate (by reference) the OCI Code of Conduct.
+
+### Governance and Releases
+This project would incorporate the Governance and Releases processes from the OCI project template: https://github.com/opencontainers/project-template.
+
+### Project Communications
+This project would continue to use existing channels in use by the OCI developer community for communication including:
+* GitHub for issues and pull requests
+* The dev@opencontainers.org email list
+* The monthly OCI developer community conference call
+* The #OpenContainers IRC channel
+
+### Versioning / Roadmap
+It aims to be a C library that can be either used as a shared library or statically linked in the projects that need it.
+
+## Frequenty Asked Questions (FAQ)
+
+**Q: Does this change the OCI Charter or Scope Table?**
+
+A: No.  Nothing in this proposal is intended to amend the OCI Charter (https://www.opencontainers.org/about/governance) or OCI Scope Table (https://www.opencontainers.org/about/oci-scope-table).


caniszczyk:
before we call a formal vote on this, I would like to get comments from @philips and the @opencontainers/runtime-spec-maintainers @opencontainers/image-spec-maintainers

proposals/libocispec.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Abstract
+Provide a C library for parsing and handle OCI runtime spec file and OCI image spec files.
+
+## Proposal
+With repositories under the http://github.com/opencontainers organization:
+
+Rename (transfer) https://github.com/giuseppe/libocispec would become https://github.com/opencontainers/libocispec.
+
+### Initial Maintainers
```
caniszczyk: any other maintainers besides you :)?

giuseppe: submitted a new version with  Qiang Huang <h.huangqiang@huawei.com> who offered to be a maintainer as well.

giuseppe:
Worth saying: I've rewritten the read_file module and licensed it Apache 2.0

README.md
```diff
@@ -24,6 +24,7 @@ https://groups.google.com/a/opencontainers.org/forum/#!forum/tob (tob@opencontai
 * [Image Format Spec](https://github.com/opencontainers/tob/tree/master/proposals/image-format)
 * [SELinux](https://github.com/opencontainers/tob/blob/master/proposals/selinux.md)
 * [Tools](https://github.com/opencontainers/tob/blob/master/proposals/tools.md)
+* [libocispec](https://github.com/opencontainers/tob/blob/master/proposals/libocispec.md)
```
wking: Previous entries in this list seem to be alphabetized, although I don't know if TOB maintainers care about that.

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
+# OCI libocispec project proposal
+
+## Abstract
+Provide a C library for parsing and handle OCI runtime spec file and OCI image spec files.
```
wking: “handle” → “handling”.  Or maybe “representing”?  Or maybe “passing”?  The library seems to handle parsing from various formats into C types and manipulating those C types in memory.  There's nothing about implementing the OCI spec semantics.  It would be good to have a short word/phrase here to make that distinction clear, and I don't think “handling” is specific enough.

Also, the input may be in a file, or may just be in a string (e.g. `oci_container_parse_data`).  What they have in common is that they all parse JSON.  So I'd rather end with “OCI Runtime and Image specification JSON.” or some such.

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Proposal
+With repositories under the http://github.com/opencontainers organization:
+
+Rename (transfer) https://github.com/giuseppe/libocispec would become https://github.com/opencontainers/libocispec.
```
wking: “would become” → “to”.

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+* The #OpenContainers IRC channel
+
+### Versioning / Roadmap
+It aims to be a C library to be statically linked in the projects that need it.
+
+## Frequenty Asked Questions (FAQ)
+
+**Q: Does this change the OCI Charter or Scope Table?**
+
+A: No.  Nothing in this proposal is intended to amend the OCI Charter (https://www.opencontainers.org/about/governance) or OCI Scope Table (https://www.opencontainers.org/about/oci-scope-table).
```
wking: Really?  There's nothing in the scope table about this.  I think we should ask for an entry covering this project.  Although previous proposals that should have impacted the scope table have had their associated scope-table updates [punted to post-proposal][1], so maybe we want to go that way again.

[1]: https://github.com/opencontainers/tob/pull/22#discussion_r93338910

tianon:
IMO this looks like a really cool project, but I think it's a bit premature for the OCI to officially adopt it, especially since it's not used by any of the other projects that the OCI already maintains.  I'm all for the OCI linking to it and reconsidering officialization when it's slightly more mature.

giuseppe:
@tianon I am not sure how more mature it can be.  If the OCI runtime/image specs won't change in a way to break the `generate.py` script then I don't foresee any major work on it.

I know that @x0916 and @tanyifeng are using it in their project (it is not publicly available AFAIK)

hqhq:
@tianon @giuseppe Yes we're using it in our internal project in Huawei, it's still premature and not open sourced yet. I think the implementation of this library is pretty straightforward and should be stable, most of the maintenance work would be adapting the changes in specs, so there won't be significantly more mature IMO.

Regarding the usage by other OCI projects, since the major user of this library would be spec implementations, I think we'll only have one implementation for one spec in OCI according to our pre-discussions, and that implementation would most likely be a go version, so I doubt there would be any c version implementation which uses this library in OCI.

rhatdan:
Any update on this?  @giuseppe is using this for a cool implementation of crun (https://github.com/giuseppe/crun), basic an OCI runtime written totally in C.

caniszczyk:
@rhatdan once the new @opencontainers/tob is formed at the end of this month, they can take a look at it.

rhatdan:
@caniszczyk Thanks.

README.md
```diff
@@ -23,6 +23,7 @@ https://groups.google.com/a/opencontainers.org/forum/#!forum/tob (tob@opencontai
 * [Digest](proposals/digest.md)
 * [Distribution Spec](proposals/distribution.md)
 * [Image Format Spec](proposals/image-format)
+* [libocispec](https://github.com/opencontainers/tob/blob/master/proposals/libocispec.md)
```
wking: Can you make this a relative link?  See #51.

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
+# OCI libocispec project proposal
+
+## Abstract
+Provide a C library for parsing and accessing OCI Runtime and Image specification JSON.
+
+## Proposal
+With repositories under the http://github.com/opencontainers organization:
```
wking: http -> https

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+
+Rename (transfer) https://github.com/giuseppe/libocispec to https://github.com/opencontainers/libocispec.
+
+### Initial Maintainers
+Initial maintainers of the libocispec project:
+* Giuseppe Scrivano <gscrivan@redhat.com> (@gscrivano)
+* Qiang Huang <h.huangqiang@huawei.com>
```
wking: (@hqhq)

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+* Daniel J Walsh <dwalsh@redhat.com> (@rhatdan)
+
+### Code of Conduct
+This project would incorporate (by reference) the OCI Code of Conduct.
+
+### Governance and Releases
+This project would incorporate the Governance and Releases processes from the OCI project template: https://github.com/opencontainers/project-template.
+
+### Project Communications
+Both of the proposed projects would continue to use existing channels in use by the OCI developer community for communication including:
```
wking: "Both of the proposed projects" -> "This project"

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+* GitHub for issues and pull requests
+* The dev@opencontainers.org email list
+* The weekly OCI developer community conference call
```
wking: "weekly" -> "monthly"

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+* The #OpenContainers IRC channel
+
+### Versioning / Roadmap
+It aims to be a C library to be statically linked in the projects that need it.
```
wking: What about versioning?  I expect the ingtersection between sopported JSON and the exposed API will be interesting.  Do you intend to buffer callers from breaking JSON schema changes?  I expect not.

giuseppe:
@wking thanks, I've addressed your comments and pushed a new version

vbatts:
I'm not opposed to libocispec. I see value in having a C implementation 
of OCI features.

Other's thoughts?

caniszczyk:
I think that's for the @opencontainers/tob to decide, you have the power to do what you want. I would put this more of a utility (ala runtime/image tools). The other issue with this project is that @giuseppe would have to relicense it under Apache v2 before it can even be considered for OCI

giuseppe:
I am fine to relicense it under Apache v2 if there is agreement on adding it to OCI.

@crosbymichael, libocispec is not an implementation of the OCI specs, it is only the C parser (automatically generated from the JSON schema).

https://github.com/giuseppe/crun is the OCI implementation that uses libocispec.

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Abstract
+Provide a C library for parsing and accessing OCI Runtime and Image specification JSON.
+
+## Proposal
+With repositories under the https://github.com/opencontainers organization:
+
+Rename (transfer) https://github.com/giuseppe/libocispec to https://github.com/opencontainers/libocispec.
+
+### Initial Maintainers
```
crosbymichael: Is there a reason why you are proposing maintainers here that have not contributed code or issues directly to the project? 

https://github.com/giuseppe/libocispec/graphs/contributors

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
+# OCI libocispec project proposal
+
+## Abstract
+Provide a C library for parsing and accessing OCI Runtime and Image specification JSON.
```
mikebrow: Provide code for parsing and accessing JSON versions of OCI Runtime and Image specs. The first implementation will be a C library automatically generated directly from the OCI Runtime and Image specifications.

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Proposal
+With repositories under the https://github.com/opencontainers organization:
+
+Rename (transfer) https://github.com/giuseppe/libocispec to https://github.com/opencontainers/libocispec.
```
mikebrow: What about moving it here:
https://github.com/opencontainers/runtime-tools   maybe in contrib/libocispec/c/

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Project Communications
+This project would continue to use existing channels in use by the OCI developer community for communication including:
+* GitHub for issues and pull requests
+* The dev@opencontainers.org email list
+* The monthly OCI developer community conference call
+* The #OpenContainers IRC channel
+
+### Versioning / Roadmap
+It aims to be a C library to be statically linked in the projects that need it.
```
mikebrow: Guess that answers my versioning question :-)

proposals/libocispec.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Proposal
+With repositories under the https://github.com/opencontainers organization:
+
+Rename (transfer) https://github.com/giuseppe/libocispec to https://github.com/opencontainers/libocispec.
+
+### Initial Maintainers
+Initial maintainers of the libocispec project:
+* Giuseppe Scrivano <gscrivan@redhat.com> (@giuseppe)
+* Qiang Huang <h.huangqiang@huawei.com> (@hqhq)
```
hqhq: Sorry I'm not available to contribute or maintain this project now, I would recommend my colleague @tanyifeng who's been contributing to and using this project, I think he would be willing and qualified to do this job : )

proposals/libocispec.md
wking: > What about moving it here:
https://github.com/opencontainers/runtime-tools maybe in contrib/libocispec/c/

My concern with that is that versioning the repo becomes tricky.  Do you lock the Go and C APIs together under a single version?  Version them separately with prefixed tags for C releases?  And you'd need to check with runtime-tools maintainers to see if enough of them care about and feel comfortable enough with C to maintain this as well.  I don't have a problem with this becoming an OCI Project in its own right, or remaining outside the OCI umbrella.  But I think stuffing it into an existing OCI Project may be more trouble than its worth.

And however this shakes out, we probably want to link the library from [here][1] to raise awareness.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/implementations.md#testing--tools

proposals/libocispec.md
giuseppe: at the time I proposed libocispec I asked around if someone was interested in co-maintaining it so to not have only one maintainer listed.  I'll update the list if this is not the preferred way of doing it.

proposals/libocispec.md
giuseppe: I guess I have to update this part :-) now libocispec supports also to be a shared library

proposals/libocispec.md
giuseppe: sure.

@tanyifeng, are you fine to be added here?

tanyifeng: Of course, I'll be glad to be a maintainer of libocispec.

giuseppe:
I've updated the proposal :arrow_up:

giuseppe:
is there still any interest in moving libocispec to opencontainers?

vbatts:
i still have interest, but for now I am not sure it makes sense, given the adoption and use. Better likely to have a page for implementations by programming language.

giuseppe:
given there is no interest, I am going to close this PR.  Thanks to anyone who partecipated in the discussion.

PR #35, opened 2018-01-23
Add distribution spec project proposal

caniszczyk:
based off initial discussion here: https://docs.google.com/document/d/15y0SBrrDFIEM7pnU-Oe3Y6pq-eTZfo0mk-k33cS2hR4/edit#

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

### DIFF
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -103,19 +103,16 @@ The following entries should be added to the [scope table][scope]:
 * “Specifying a distribution method”.
     This entry replaces part of the previous “Creating Reference spec for optional DNS based naming & distribution” and “Standardizing on a particular Distribution method” entries.
 
-    Retrieving image indexes covers the current “tag listing” (e.g. “what named manifests are in `library/busybox`?”), because tags are entries in the image format's [`manifests` array][manifests].
-    Other tag-listing endpoints needed for backwards-compatibility are therefore in scope as well.
+    Tag-listing (e.g. “what named manifests are in `library/busybox`?”) endpoints are in scope and required for backwards compatibility with clients.
+    An endpoint for listing image repository names within a registry is considered out of scope for this specification.
 
-    Grouping image indexes in repositories is considered part of distribution policy or content management, which are out of scope for this entry's per-image action.
-    For example, “what images are under `library/`?” is out of scope for this project.
+    Managing the grouping of image repository names is considered part of distribution policy or content management, which are out of scope.
+    For example, “which image repositories are under `library/`?” is out of scope for this project.
 
     * What: Specifying a distribution method
     * In/Out/Future: In scope
     * Status: In progress (see opencontainers/distribution-spec)
-    * Description: Define a protocol for creating, retrieving, updating, and deleting objects defined in the [image specification][image-spec].
-        Listing repositories (like [`/v2/_catalog`][catalog]) is a multi-[image-index][] action, which is out of scope for this entry.
-        Managing groups of image indexes requires multi-[image-index][] actions, which are out of scope for this entry.
-        Listing image indexes within a group is a multi-[image-index][] action, which is out of scope for this entry.
+    * Description: Define a protocol for creating, retrieving, updating, and deleting content-addressable objects, such as those defined in the [image specification][image-spec].
     * Why: This specification will provide one (of possibly many) distribution specifications.
         Alternative distribution specifications may be developed for uses cases not covered by this specification, but defining them is currently out of scope for the OCI.
 
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -1,0 +1,61 @@
+# Abstract
+
+The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/index.md](https://github.com/docker/distribution/blob/master/docs/spec/index.md)).
+
+In the OCI, having a solid, common distribution specification with conformance testing will ensure long lasting security and interoperability throughout the container ecosystem.
+
+## Proposal
+
+TL;DR; Move [https://github.com/docker/distribution/tree/master/docs/spec](https://github.com/docker/distribution/tree/master/docs/spec) to [https://github.com/opencontainers/distribution-spec](https://github.com/opencontainers/distribution-spec)
+
+This proposal covers the distribution API spec, and while it does not cover the code for the docker-registry, that implementation is considered the reference implementation. There are other implementations of this protocol, not all are open-source though (Google gcr.io, Amazon ECR, CoreOS Quay, Gitlab registry, JFrog Artifactory registry, Huawei Dockyard, etc).
+
+In the past when the topic of having an OCI specification around the distribution of container images, it was deferred as "let’s get the image format defined, mean while the industry will settle on a distribution standard". Fast forward, OCI image format is out and adopted, and the Registry v2 is the defacto standard. There is and will be use-cases for alternate methods and the future will likely hold creative ways to push, fetch and share container images, but right now this promotion serves to acknowledge by the OCI the current industry standard of distributing container images.
+
+There is polish that is needed e.g. broken links to storage-driver docs, as well as making sections more generic regarding the OCI descriptors and media-types, but on the whole this is a lateral move.
+
+## Initial Maintainers
+
+* Stephen Day <stephen.day@docker.com> (@stevvooe)
+* Vincent Batts <vbatts@redhat.com> (@vbatts)
+* Derek McGowan <derek.mcgowan@docker.com> (@dmcgowan)
+
+Additional Maintainers to consider:
+
+* Ahmet Alp Balkan (Google)
+* Matt Moore (Google)
+* Yuwa (MSFT)
+* Clayton Coleman (Red Hat)
+* Antonio Murdaca (@runcom) (Red Hat)
+* Samuel Karp (@samuelkarp) (AWS)
+* Mike Brown (IBM)
+* Jimmy Zelinskie jimmy@coreos.com (@jzelinskie)
+* Liu Genping <[liugenping@huawei.com](mailto:liugenping@huawei.com)>
+* Vanessa Sochat (@vsoch)
+
+## Code of Conduct
+
+This project would incorporate (by reference) the OCI Code of Conduct ([https://github.com/opencontainers/tob/blob/master/code-of-conduct.md](https://github.com/opencontainers/tob/blob/master/code-of-conduct.md)).
+
+## Governance and Releases
+
+This project would incorporate the Governance and Releases processes from the OCI project template: [https://github.com/opencontainers/project-template](https://github.com/opencontainers/project-template).
+
+## Project Communications
+
+Both of the proposed projects would continue to use existing channels in use by the OCI developer community for communication including:
+
+* GitHub for issues and pull requests
+* The dev@opencontainers.org email list
+* The monthly OCI developer community conference call
+* The #OpenContainers freenode IRC channel
+
+## Versioning / Roadmap
+
+The API spec is currently considered v2 and we will start the specification at v2.0. Fewer places to change and compare, and it would keep with it being a lateral move.
+
+## Frequently Asked Questions (FAQ)
+
+* Does this include the code of the docker-registry?
+    * No. This is an API specification discussion.
+
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -85,9 +85,8 @@ The following entries should be added to the [scope table][scope]:
     * In/Out/Future: Out of scope
     * Status: N/A
     * Description: Defining protocols for authenticating and authorizing distribution access.
-    * Why: As a HTTP-based protocol, clients and servers can negotiate authentication via HTTP's [challenge-response authentication framework][rfc7235-s2.1].
+    * Why: As an HTTP-based protocol, clients and servers can negotiate authentication via HTTP's [challenge-response authentication framework][rfc7235-s2.1].
         There is no need for the distribution specification to choose a particular authentication scheme, because clients receiving 401 and 407 responses can use IANA's [HTTP Authentication Scheme Registry][iana-auth] to look up referenced schemes and take appropriate action.
-        It is reasonable to provide a standardized way to use DNS based distribution in conjunction with OCI without requiring its use.
 
 * “Creating a reference spec for optional DNS based naming and discovery”.
     Discovery and registry protocols are completely separate and do not need to be added together.
@@ -95,9 +94,9 @@ The following entries should be added to the [scope table][scope]:
 
     * What: Creating a reference spec for optional DNS based naming and discovery
     * In/Out/Future: In scope for future specification
-    * Status: Not currently being worked
+    * Status: Work not yet started
     * Description: Define a protocol for resolving an image name to retrieval information.
-        When we address this, we will also allow for alternative, parallel name-to-image discovery protocols in parallel with the OCI-specified protocol.
+        When we address this, we will also allow for alternative name-to-image discovery protocols in parallel with the OCI-specified protocol.
     * Why: It is reasonable to provide a standardized way to use DNS based distribution in conjunction with OCI without requiring its use.
         There are many good use cases for DNS based distribution, but not all use cases support this.
         Furthermore, encoding the location of a bundle into the bundle can cause issues with downloads from alternate locations other than the origin specified in the name.
@@ -123,13 +122,13 @@ The following entries should be added to the [scope table][scope]:
 
 * “Retrieving images by their content-addressable hash”.
     Docker's registery API already provides endpoints for fetching manifest objects by digest][get-manifest].
-    Docker's registery API does not currently provides endpoints for fetching image objects by digest, but this is the project where that will happen.
+    Docker's registery API does not currently provide endpoints for fetching image objects by digest, but this is the project where that will happen.
 
     * What: Retrieving images by their content-addressable hash
     * In/Out/Future: In scope
     * Status: In progress (see opencontainers/distribution-spec)
     * Description: Specify a protocol for retrieving an image from a distribution engine by the image's content-addressable hash.
-    * Why: Using a hash as a name is a way to ensure a unique image name without relying on a particular naming authority/or system.
+    * Why: Using a hash as a name is a way to ensure a unique image name without relying on a particular naming authority or system.
         Using hashing for name is an acceptable addition as it does not encode any centralized namespace.
 
 The following entries should remain in the [scope table][scope] but not be addressed by this project:
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -86,7 +86,6 @@ The following entries should be added to the [scope table][scope]:
     * Status: N/A
     * Description: Defining protocols for authenticating and authorizing distribution access.
     * Why: As an HTTP-based protocol, clients and servers can negotiate authentication via HTTP's [challenge-response authentication framework][rfc7235-s2.1].
-        There is no need for the distribution specification to choose a particular authentication scheme, because clients receiving 401 and 407 responses can use IANA's [HTTP Authentication Scheme Registry][iana-auth] to look up referenced schemes and take appropriate action.
 
 * “Creating a reference spec for optional DNS based naming and discovery”.
     Discovery and registry protocols are completely separate and do not need to be added together.
@@ -144,7 +143,6 @@ The following entries should remain in the [scope table][scope] but not be addre
 [api.md]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md
 [catalog]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#catalog
 [get-manifest]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#pulling-an-image-manifest
-[iana-auth]: http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
 [image-spec]: https://github.com/opencontainers/image-spec/
 [image-index]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md
 [manifest]: https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -1,16 +1,17 @@
 # Abstract
 
-The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/api.md](https://github.com/docker/distribution/blob/master/docs/spec/api.md)).
+The Docker registry protocol has become the defacto standard across the container registry world.
 
 In the OCI, having a solid, common distribution specification with conformance testing will ensure long lasting security and interoperability throughout the container ecosystem.
 
 ## Proposal
 
-TL;DR; Move [https://github.com/docker/distribution/tree/master/docs/spec](https://github.com/docker/distribution/tree/master/docs/spec) to [https://github.com/opencontainers/distribution-spec](https://github.com/opencontainers/distribution-spec)
+TL;DR; Move [`api.md`][api.md] to a new [distribution-spec project](https://github.com/opencontainers/distribution-spec).
 
 This proposal covers the distribution API spec, and while it does not cover the code for the docker-registry, that implementation is considered the reference implementation. There are other implementations of this protocol, not all are open-source though (Google gcr.io, Amazon ECR, CoreOS Quay, Gitlab registry, JFrog Artifactory registry, Huawei Dockyard, etc).
 
 In the past when the topic of having an OCI specification around the distribution of container images was discussed, it was deferred as "let’s get the image format defined, meanwhile the industry will settle on a distribution standard". Fast forward, OCI image format is out and adopted, and the Registry v2 is the defacto standard. There is and will be use-cases for alternate methods and the future will likely hold creative ways to push, fetch and share container images, but right now this promotion serves to acknowledge by the OCI the current industry standard of distributing container images.
+This proposal also provides the container ecosystem with a means to discuss and schedule extensions to the distribution specification.
 
 There is polish that is needed e.g. broken links to storage-driver docs, as well as making sections more generic regarding the OCI descriptors and media-types, but on the whole this is a lateral move.
 
@@ -57,10 +58,96 @@ The API spec is currently considered v2 and we will start the specification at v
 
 ## Frequently Asked Questions (FAQ)
 
-* Does this include the code of the docker-registry?
-    * No. This is an API specification discussion.
+**Q: Does this include the code of the docker-registry?**
+
+A: No. This is an API specification discussion.
+
+**Q: Does this change the OCI Charter or Scope Table?**
+
+A: Not the charter, but it does change the scope table.
+This project is scoped to specifying per-image client ↔ registry interaction with an [HTTP][rfc7230]-based protocol.
+The following scope entries should be removed from the [scope table][scope]:
+
+* “Use of Hash as Content Addressable name for immutable containers”.
+    This entry is in scope for this project, and a more detailed entry will be added as described below.
+* “Creating Reference spec for optional DNS based naming & distribution”.
+    This entry conflates naming and distribution, which will be separated by this proposal.
+* “Standardizing on a particular Distribution method”.
+    This proposal will provide one (of possibly many) distribution specifications, so the old “There is no current agreement on how to distribute content” no longer applies.
+
+The following entries should be added to the [scope table][scope]:
+
+* “Specifying authentication and authorization schemes”.
+    Docker's current registry uses an [extension][token] of the [`Bearer`][rfc6750] [auth scheme][rfc7235-s2.1].
+    Work on specifying Docker's scheme will continue independently, and is orthogonal to the registry API.
+
+    * What: Specifying authentication and authorization schemes
+    * In/Out/Future: Out of scope
+    * Status: N/A
+    * Description: Defining protocols for authenticating and authorizing distribution access.
+    * Why: As a HTTP-based protocol, clients and servers can negotiate authentication via HTTP's [challenge-response authentication framework][rfc7235-s2.1].
+        There is no need for the distribution specification to choose a particular authentication scheme, because clients receiving 401 and 407 responses can use IANA's [HTTP Authentication Scheme Registry][iana-auth] to look up referenced schemes and take appropriate action.
+        It is reasonable to provide a standardized way to use DNS based distribution in conjunction with OCI without requiring its use.
+
+* “Creating a reference spec for optional DNS based naming and discovery”.
+    Discovery and registry protocols are completely separate and do not need to be added together.
+    This entry replaces part of the previous “Creating Reference spec for optional DNS based naming & distribution” entry.
+
+    * What: Creating a reference spec for optional DNS based naming and discovery
+    * In/Out/Future: In scope for future specification
+    * Status: Not currently being worked
+    * Description: Define a protocol for resolving an image name to retrieval information.
+        When we address this, we will also allow for alternative, parallel name-to-image discovery protocols in parallel with the OCI-specified protocol.
+    * Why: It is reasonable to provide a standardized way to use DNS based distribution in conjunction with OCI without requiring its use.
+        There are many good use cases for DNS based distribution, but not all use cases support this.
+        Furthermore, encoding the location of a bundle into the bundle can cause issues with downloads from alternate locations other than the origin specified in the name.
+
+* “Specifying a distribution method”.
+    This entry replaces part of the previous “Creating Reference spec for optional DNS based naming & distribution” and “Standardizing on a particular Distribution method” entries.
+
+    Retrieving images covers the current “tag listing” (e.g. “what named manifests are in `library/busybox`?”), because tags are entries in the image format's [`manifests` array][manifests].
+    Other tag-listing endpoints needed for backwards-compatibility are therefore in scope as well.
+
+    Repository actions and listing images within a repository are considered part of distribution policy or content management and are out of scope for this entry's per-image action.
+    For example, “what images are under `library/`?” is out of scope for this project.
+
+    * What: Specifying a distribution method
+    * In/Out/Future: In scope
+    * Status: In progress (see opencontainers/distribution-spec)
+    * Description: Define a protocol for image, manifest, config, and blob creation, retrieval, and deletion.
+        Listing repositories is a multi-repository action, which is out of scope for this entry.
+        Creating and deleting repositories are per-repository actions, which are out of scope for this entry.
+        Listing images within repositories is a per-repository action, which is out of scope for this entry.
+    * Why: This specification will provide one (of possibly many) distribution specifications.
+        Alternative distribution specifications may be developed for uses cases not covered by this specification, but defining them is currently out of scope for the OCI.
+
+* “Retrieving images by their content-addressable hash”.
+    Docker's registery API already provides endpoints for fetching manifest objects by digest][get-manifest].
+    Docker's registery API does not currently provides endpoints for fetching image objects by digest, but this is the project where that will happen.
+
+    * What: Retrieving images by their content-addressable hash
+    * In/Out/Future: In scope
+    * Status: In progress (see opencontainers/distribution-spec)
+    * Description: Specify a protocol for retrieving an image from a distribution engine by the image's content-addressable hash.
+    * Why: Using a hash as a name is a way to ensure a unique image name without relying on a particular naming authority/or system.
+        Using hashing for name is an acceptable addition as it does not encode any centralized namespace.
+
+The following entries should remain in the [scope table][scope] but not be addressed by this project:
+
+* “Specifying way to attach signatures”.
+    We don't need to address this as part of distribution, because all resources are content-addressable and can be signed in external systems.
 
 ## Related GitHub Issues
 
 * Simplifies tag listing: docker/distribution#2169
 * Allows listing of manifests: docker/distribution#2199
+
+[api.md]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md
+[get-manifest]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#pulling-an-image-manifest
+[iana-auth]: http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
+[manifests]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md#L23
+[rfc6750]: https://tools.ietf.org/html/rfc6750
+[rfc7230]: https://tools.ietf.org/html/rfc7230
+[rfc7235-s2.1]: https://tools.ietf.org/html/rfc7235#section-2.1
+[scope]: https://www.opencontainers.org/about/oci-scope-table
+[token]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/auth/token.md
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -32,6 +32,7 @@ Additional Maintainers to consider:
 * Jimmy Zelinskie jimmy@coreos.com (@jzelinskie)
 * Liu Genping <[liugenping@huawei.com](mailto:liugenping@huawei.com)>
 * Vanessa Sochat (@vsoch) (Stanford) <vsochat@stanford.edu>
+* Eduardo Arango (@ArangoGutierrez) (Sylabs) <eduardo@sylabs.io>
 
 ## Code of Conduct
 
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -1,6 +1,6 @@
 # Abstract
 
-The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/index.md](https://github.com/docker/distribution/blob/master/docs/spec/index.md)).
+The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/api.md](https://github.com/docker/distribution/blob/master/docs/spec/api.md)).
 
 In the OCI, having a solid, common distribution specification with conformance testing will ensure long lasting security and interoperability throughout the container ecosystem.
 
@@ -10,7 +10,7 @@ TL;DR; Move [https://github.com/docker/distribution/tree/master/docs/spec](https
 
 This proposal covers the distribution API spec, and while it does not cover the code for the docker-registry, that implementation is considered the reference implementation. There are other implementations of this protocol, not all are open-source though (Google gcr.io, Amazon ECR, CoreOS Quay, Gitlab registry, JFrog Artifactory registry, Huawei Dockyard, etc).
 
-In the past when the topic of having an OCI specification around the distribution of container images, it was deferred as "let’s get the image format defined, mean while the industry will settle on a distribution standard". Fast forward, OCI image format is out and adopted, and the Registry v2 is the defacto standard. There is and will be use-cases for alternate methods and the future will likely hold creative ways to push, fetch and share container images, but right now this promotion serves to acknowledge by the OCI the current industry standard of distributing container images.
+In the past when the topic of having an OCI specification around the distribution of container images was discussed, it was deferred as "let’s get the image format defined, meanwhile the industry will settle on a distribution standard". Fast forward, OCI image format is out and adopted, and the Registry v2 is the defacto standard. There is and will be use-cases for alternate methods and the future will likely hold creative ways to push, fetch and share container images, but right now this promotion serves to acknowledge by the OCI the current industry standard of distributing container images.
 
 There is polish that is needed e.g. broken links to storage-driver docs, as well as making sections more generic regarding the OCI descriptors and media-types, but on the whole this is a lateral move.
 
@@ -31,7 +31,7 @@ Additional Maintainers to consider:
 * Mike Brown (IBM)
 * Jimmy Zelinskie jimmy@coreos.com (@jzelinskie)
 * Liu Genping <[liugenping@huawei.com](mailto:liugenping@huawei.com)>
-* Vanessa Sochat (@vsoch)
+* Vanessa Sochat (@vsoch) (Stanford) <vsochat@stanford.edu>
 
 ## Code of Conduct
 
@@ -59,3 +59,7 @@ The API spec is currently considered v2 and we will start the specification at v
 * Does this include the code of the docker-registry?
     * No. This is an API specification discussion.
 
+## Related GitHub Issues
+
+* Simplifies tag listing: docker/distribution#2169
+* Allows listing of manifests: docker/distribution#2199
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -104,30 +104,30 @@ The following entries should be added to the [scope table][scope]:
 * “Specifying a distribution method”.
     This entry replaces part of the previous “Creating Reference spec for optional DNS based naming & distribution” and “Standardizing on a particular Distribution method” entries.
 
-    Retrieving images covers the current “tag listing” (e.g. “what named manifests are in `library/busybox`?”), because tags are entries in the image format's [`manifests` array][manifests].
+    Retrieving image indexes covers the current “tag listing” (e.g. “what named manifests are in `library/busybox`?”), because tags are entries in the image format's [`manifests` array][manifests].
     Other tag-listing endpoints needed for backwards-compatibility are therefore in scope as well.
 
-    Repository actions and listing images within a repository are considered part of distribution policy or content management and are out of scope for this entry's per-image action.
+    Grouping image indexes in repositories is considered part of distribution policy or content management, which are out of scope for this entry's per-image action.
     For example, “what images are under `library/`?” is out of scope for this project.
 
     * What: Specifying a distribution method
     * In/Out/Future: In scope
     * Status: In progress (see opencontainers/distribution-spec)
-    * Description: Define a protocol for image, manifest, config, and blob creation, retrieval, and deletion.
-        Listing repositories is a multi-repository action, which is out of scope for this entry.
-        Creating and deleting repositories are per-repository actions, which are out of scope for this entry.
-        Listing images within repositories is a per-repository action, which is out of scope for this entry.
+    * Description: Define a protocol for creating, retrieving, updating, and deleting objects defined in the [image specification][image-spec].
+        Listing repositories (like [`/v2/_catalog`][catalog]) is a multi-[image-index][] action, which is out of scope for this entry.
+        Managing groups of image indexes requires multi-[image-index][] actions, which are out of scope for this entry.
+        Listing image indexes within a group is a multi-[image-index][] action, which is out of scope for this entry.
     * Why: This specification will provide one (of possibly many) distribution specifications.
         Alternative distribution specifications may be developed for uses cases not covered by this specification, but defining them is currently out of scope for the OCI.
 
-* “Retrieving images by their content-addressable hash”.
-    Docker's registery API already provides endpoints for fetching manifest objects by digest][get-manifest].
-    Docker's registery API does not currently provide endpoints for fetching image objects by digest, but this is the project where that will happen.
+* “Retrieving image content by its content-addressable hash”.
+    Docker's registry API already provides [endpoints for fetching manifest objects by digest][get-manifest].
+    Docker's registry API does not currently provide endpoints for fetching [image-index][] objects by digest, but this is the project where that will happen.
 
-    * What: Retrieving images by their content-addressable hash
+    * What: Retrieving image content by its content-addressable hash
     * In/Out/Future: In scope
     * Status: In progress (see opencontainers/distribution-spec)
-    * Description: Specify a protocol for retrieving an image from a distribution engine by the image's content-addressable hash.
+    * Description: Specify a protocol for retrieving an [image index][image-index], [manifest][], or other [image specification][image-spec] object from a distribution engine by its content-addressable hash.
     * Why: Using a hash as a name is a way to ensure a unique image name without relying on a particular naming authority or system.
         Using hashing for name is an acceptable addition as it does not encode any centralized namespace.
 
@@ -142,8 +142,12 @@ The following entries should remain in the [scope table][scope] but not be addre
 * Allows listing of manifests: docker/distribution#2199
 
 [api.md]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md
+[catalog]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#catalog
 [get-manifest]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#pulling-an-image-manifest
 [iana-auth]: http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
+[image-spec]: https://github.com/opencontainers/image-spec/
+[image-index]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md
+[manifest]: https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md
 [manifests]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md#L23
 [rfc6750]: https://tools.ietf.org/html/rfc6750
 [rfc7230]: https://tools.ietf.org/html/rfc7230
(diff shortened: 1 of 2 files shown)


proposals/distribution.md
```diff
@@ -0,0 +1,61 @@
...
+
+There is polish that is needed e.g. broken links to storage-driver docs, as well as making sections more generic regarding the OCI descriptors and media-types, but on the whole this is a lateral move.
+
+## Initial Maintainers
+
+* Stephen Day <stephen.day@docker.com> (@stevvooe)
+* Vincent Batts <vbatts@redhat.com> (@vbatts)
+* Derek McGowan <derek.mcgowan@docker.com> (@dmcgowan)
+
+Additional Maintainers to consider:
```
wking: The TOB needs to pick an initial maintainer set.  I think these suggestions are for the TOB, and should be reviewed and either added to the initial maintainer list or dropped before the vote.  A three-person initial maintainer set can add other maintainers on their own, so the TOB can drop tge whole consideration list and punt to the initial maintainers if it wants.

proposals/distribution.md
```diff
@@ -0,0 +1,61 @@
+# Abstract
+
+The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/index.md](https://github.com/docker/distribution/blob/master/docs/spec/index.md)).
```
wking: Pin this link to a specific version?  Or does Docker intend to keep docs in that master after this proposal?

proposals/distribution.md
```diff
@@ -0,0 +1,61 @@
...
+
+This proposal covers the distribution API spec, and while it does not cover the code for the docker-registry, that implementation is considered the reference implementation. There are other implementations of this protocol, not all are open-source though (Google gcr.io, Amazon ECR, CoreOS Quay, Gitlab registry, JFrog Artifactory registry, Huawei Dockyard, etc).
+
+In the past when the topic of having an OCI specification around the distribution of container images, it was deferred as "let’s get the image format defined, mean while the industry will settle on a distribution standard". Fast forward, OCI image format is out and adopted, and the Registry v2 is the defacto standard. There is and will be use-cases for alternate methods and the future will likely hold creative ways to push, fetch and share container images, but right now this promotion serves to acknowledge by the OCI the current industry standard of distributing container images.
```
duglin: I think you might be missing a `was discussed` before the first comma (...of container images *was discussed*,)

duglin: s/mean while/meanwhile/

duglin:
minor typos, but LGTM

proposals/distribution.md
```diff
@@ -0,0 +1,61 @@
...
+
+This project would incorporate the Governance and Releases processes from the OCI project template: [https://github.com/opencontainers/project-template](https://github.com/opencontainers/project-template).
+
+## Project Communications
+
+Both of the proposed projects would continue to use existing channels in use by the OCI developer community for communication including:
+
+* GitHub for issues and pull requests
+* The dev@opencontainers.org email list
+* The monthly OCI developer community conference call
```
wking: Does starting at v2 (as suggested below) get us out of the [pre-1.0 weekly meeting][1] recommendation?  I'm not sure if the "we're starting at v2" approach was considered when writing those docs (I certainly hadn't considered it).

[1]: https://github.com/opencontainers/project-template/blame/8e4c7e3dbea403c727d7fc526b8dcd53d6c98202/RELEASES.md#L34

proposals/distribution.md
```diff
@@ -0,0 +1,61 @@
...
+* Matt Moore (Google)
+* Yuwa (MSFT)
+* Clayton Coleman (Red Hat)
+* Antonio Murdaca (@runcom) (Red Hat)
+* Samuel Karp (@samuelkarp) (AWS)
+* Mike Brown (IBM)
+* Jimmy Zelinskie jimmy@coreos.com (@jzelinskie)
+* Liu Genping <[liugenping@huawei.com](mailto:liugenping@huawei.com)>
+* Vanessa Sochat (@vsoch)
+
```
vsoch: For completeness, choose any/all that is needed:
 - affiliation: Stanford / Singularity
 - email: vsochat@stanford.edu

proposals/distribution.md
stevvooe: This is actually the wrong link. The specification is at https://github.com/docker/distribution/blob/master/docs/spec/api.md. The auth and token spec are docker-specific extensions and are not required for a full specification.

stevvooe:
This should probably also include two active PRs:

- Simplifies tag listing: https://github.com/docker/distribution/pull/2169
- Allows listing of manifests: https://github.com/docker/distribution/pull/2199

ArangoGutierrez:
+1

proposals/distribution.md
vsoch: Thanks!

proposals/distribution.md
```diff
@@ -0,0 +1,65 @@
+# Abstract
+
+The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/api.md](https://github.com/docker/distribution/blob/master/docs/spec/api.md)).
```
wking: I [still][1] think we want to pin this to a specific version.  For example:

    https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md

[1]: https://github.com/opencontainers/tob/pull/35#discussion_r163291623

wking:
> This should probably also include two active PRs…

Do you mean “we'll want to land these in docker/distribution before the transfer” or “we'll want to open new PRs post-transfer to continue these open disussions”?  If the former, that seems like “we're not ready to raise this proposal for a TOB vote yet”.  If the latter, that seems like something the new distribution-API maintainers can sort out amongst themselves without TOB intervention.

proposals/distribution.md
crosbymichael: Same.  

How are TOB members supposed to interpret this list?  Maintainers are picked based on current and past contributions to a codebase or area of expertise and usually their maintainer vote comes with justification of why they should be a maintainer in a project.

proposals/distribution.md
```diff
@@ -0,0 +1,66 @@
...
+
+In the OCI, having a solid, common distribution specification with conformance testing will ensure long lasting security and interoperability throughout the container ecosystem.
+
+## Proposal
```
cyphar: One thing I would like to clarify is whether we will actually be making improvements to the spec over time (as per how we did things in the other specifications) or whether we're just going to freeze on the current v2.0 and do nothing much afterwards -- since according to SemVer we'd need to work on v3.0 for that.

Also, it was my understanding that the submission of Docker Distribution as a specification would be something like "any future distribution spec we agree upon will have support for Docker Distribution" -- so as to unblock the distribution problem without locking us into never making significant progress on Distribution.

proposals/distribution.md
stevvooe: That should be an implementation detail of the proposal. The implementors of the proposal should choose the version to adopt so this can be changed without TOB involvement.

proposals/distribution.md
stevvooe: @cyphar I don't think there is anything in this protocol or proposal preventing "significant progress". The specification has been out and in the open since its inception and we have yet to receive any sizable improvement proposals. Most of the perceived limitations of this protocol are limitations in the client implementation that is part of docker.

There are some good changes that we can make in backwards compatible ways:

1. The listing PRs I mentioned here.
1. Stateless transfer. There are only a few small changes required to make this work.
2. p2p already can work on this protocol.

In addition, to this, there is a lot more that can be integrated on top of this protocol as part of client resolution. Most of the issues around naming, signing and mirroring can be taken care of with no API changes.

However, I think it should be clear that the initial specification activity should be about ensuring that the specification matches the current state of the world without breaking compatibility.

proposals/distribution.md
AkihiroSuda: I think the problem is that the hyperlink will be broken when the spec is moved to the OCI distribution-spec repo.

I think keeping the hyperlink permanently after the migration is good for recording the history.

So +1 for pinning.

proposals/distribution.md
AkihiroSuda: It was discussed in the google doc

proposals/distribution.md
vsoch: I think what would be good to see is some clear instructions for the non-expert in terms of writing and working on specifications. Coming from this kind of boat, I have the general sentiment that I want to contribute, and I want to learn about the process to improve expertise in this sort of contribution. I would look for something akin to a CONTRIBUTING.md, or even a simple bullet list in a README that goes through logical steps. Like:

 - Goal: to resolve conflict for technical specifications
 - Procedure:
 1. receiive / link to protocol for discussion
 2. board comments and asks questions as issues
 3. assignment goes to X
 4. issued discussed and resolved, something else...
 5. head maintainer(s) have final sign off on something
 6. publication via... X

I suppose this is a technical standard for reviewing technical standards, haha. For example, a nice parallel is to look at something like [JOSS](http://joss.theoj.org/) that has complete orchestration via a little robot integration. The workflow is clearly defined for the reviewer and reviewee, and the assigned editor. The parties involved fill in the gaps in terms of opening and resolving issues, but there is always clear definition. I'm not sure how something like this might fit for our group, but it would be great to think about to get greater community involvement and discussion.

proposals/distribution.md
wking: And as long as the TOB-selected version isn't far behind (how far will the spec move during a week of voting?), it should be easy for the new maintainets to catch up on any subsequent drift.

proposals/distribution.md
wking: > The workflow is clearly defined for the reviewer and reviewee, and the assigned editor.

project-template (referenced from this proposal) is what the OCI ises for this role.  See especially [here][1] and [here][2].

[1]: https://github.com/opencontainers/project-template/blob/master/CONTRIBUTING.md
[2]: https://github.com/opencontainers/project-template/blob/master/MAINTAINERS_GUIDE.md

proposals/distribution.md
vsoch: hey @wking here you go! If you follow the link at the top, click on "Comments" in the top right, you can explore some of the discussion.

image

proposals/distribution.md
vsoch: Thank @wking ! The general roles for a contributor and maintainer are well defined, but I'm wondering about a more specific "hand holding" how to contribute sort of document, even just a bullet list. If it's something that can be observed then I can observe, learn, and write something up. I'm mostly thinking about this to help new contributors such as myself to get into the ropes of the group :)

wking: > I'm wondering about a more specific "hand holding" how to contribute sort of document, even just a bullet list.

"How to contribute" is exactly what `CONTRIBUTING.md` tries to cover.  What do you feel us m8ssing?  Perhaps file project-template issues / PRs to avoid drift here?

vsoch: Sure thing! A specific question before closing the discussion here. For this PR, given that we don't have code (nothing to test or critique), and it's a general text draft, what are the criteria we are using for evaluating it, and how do we know when it's ready to go? I think this is likely something that would be obvious for someone working on a lot of these drafts over time, and I'll just catch on, and it would be helpful if someone could jot down a few notes about these questions.

wking: > For this PR, given that we don't have code (nothing to test or critique), and it's a general text draft, what are the criteria we are using for evaluating it, and how do we know when it's ready to go?

Ah, *this* repo.  I thought you meant the coming distribution-spec repo.  This repo could use a `CONTRIBUTING.md`, but the voting semantics are covered in the OCI charter (linked from [here][1]).  At some point a TOB member will think it ready and put it up for a TOB vote.  Until then, we're free to suggest improvements.

[1]: https://github.com/opencontainers/tob/tree/93c9b0edefee38921f0c55526564b9b637384004#voting

cyphar: @stevvooe 

Right. Can you expand on these points (we had a short chat a while ago about them, but I don't think I fully understood what the plan was):

> * Stateless transfer. There are only a few small changes required to make this work.
> * p2p already can work on this protocol.

Also, I would still like to have some sort of `.well-known` identifier for registries so that you don't need to host a separate subdomain to have a registry. This would also be a good place to pin future extensions if it turns out that we do want to add something.

Just on this point:

> The specification has been out and in the open since its inception and we have yet to receive any sizable improvement proposals.

The reason for this may be unrelated to whether people want to make changes, or have worthwhile improvements. As you mentioned, a lot of the percieved issues with Distribution are actually because the Docker client doesn't expose those features -- and so people who want to improve distribution may not want to go through improving all layers of the Docker stack to do so.

I guess what I'm saying is "let's not close the door on any improvement discussions", especially since this is now going to be an OCI spec and not a Docker one anymore.

proposals/distribution.md
```diff
@@ -0,0 +1,66 @@
...
+## Versioning / Roadmap
+
+The API spec is currently considered v2 and we will start the specification at v2.0. Fewer places to change and compare, and it would keep with it being a lateral move.
+
+## Frequently Asked Questions (FAQ)
+
+* Does this include the code of the docker-registry?
+    * No. This is an API specification discussion.
+
+## Related GitHub Issues
```
yuwaMSFT2: What about the proposal of new APIs to the spec? Shall we do it after the vote as future incremental improvements, or shall we raise and discuss them now?

From our experience of running our registry service (Azure Container Registry), there are quite some popular feedbacks/lessons we learned and would like to see if they could be included in the future registry spec:
1. Richer image management ability, including delete repository, purge unreferenced blobs, etc.
2. Richer discovery ability where user has lots of images, like list tag by creation time, list manifest by creation time, etc. So basically to support more query parameters in the list APIs.
3. Richer metadata APIs. Basically user wants to put additional metadata with their image, in additional to a simple tag.
4. Native multi-tenant support. Something like /v2/tenants/{tenant}/repositories/{name}/manifests/{reference}. Since more and more public cloud services are adopting the registry service, a native support for multi-tenant would be great IMHO.

We would like to learn if there is any opportunity that these feedback could be included in the distribution spec. Thanks!

proposals/distribution.md
```diff
@@ -0,0 +1,61 @@
...
+
+The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/index.md](https://github.com/docker/distribution/blob/master/docs/spec/index.md)).
+
+In the OCI, having a solid, common distribution specification with conformance testing will ensure long lasting security and interoperability throughout the container ecosystem.
+
+## Proposal
+
+TL;DR; Move [https://github.com/docker/distribution/tree/master/docs/spec](https://github.com/docker/distribution/tree/master/docs/spec) to [https://github.com/opencontainers/distribution-spec](https://github.com/opencontainers/distribution-spec)
+
+This proposal covers the distribution API spec, and while it does not cover the code for the docker-registry, that implementation is considered the reference implementation. There are other implementations of this protocol, not all are open-source though (Google gcr.io, Amazon ECR, CoreOS Quay, Gitlab registry, JFrog Artifactory registry, Huawei Dockyard, etc).
```
yuwaMSFT2: Is authentication in the scope of the image distribution?

yuwaMSFT2: Also, shall we call out image signing (trusted image)?

vsoch: Image signing would be very useful for Singularity containers.

proposals/distribution.md
vsoch: Image size and format would also be very useful, beyond sniffing with a `HEAD` request. There is also a distinction between a tag (e.g., latest) and a version (a hash or commit), we've had users asking for both with Singularity Hub/Registry, and then for all these requests, the default should be reasonable of course! Are we going with docker defaults like `latest` for tag, `library` for namespace, and other registries should follow suit? @yuwaMSFT2 for tenant would that be akin to the manifests list where you can request a particular OS or architecture?  The user client would then always be required to make two calls https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list (and this is what we are doing currently in Singularity)!

proposals/distribution.md
wking: I'm fine with starting at v2.  I'm just not clear on whether monthly meetings are the right target level for a new spec (regardless of the number we put on our initial release).  The discussion in this PR suggests folks are going to have lots of ideas.  Project-template suggests weekly meetings during pre-1.0 development (presumably because the roadmap is less obvious/established then).  And there's no reason that roadmap discussions and such couldn't happen asynchronously (but then maybe we want to drop the meeting recommendation from project-template?).  Anyhow, I'd just like to make sure that, if this project starts out with monthly meetings, it was a conscious decision that we didn't expect to need weekly meetings.

proposals/distribution.md
mikebrow: we can use such as language when pointing to a particular version...

proposals/distribution.md
mikebrow: @cyphar what sort of language are you looking for to ensure the door is not closed for improvement discussions?  

How about after the interoperability statement adding:
```
This OCI distribution specification will also provide the ecosystem with a means to discuss and schedule extensions to the distribution specification.
```

proposals/distribution.md
mikebrow: @wking sounds like something to discuss on the next monthly meeting :-)

proposals/distribution.md
mikebrow: @yuwaMSFT2 @vsoch interesting extensions. See above proposal to add a sentence about discussing and scheduling in extensions without going into specific detail.

wking:
nit: This proposal needs a README entry [here][1].

[1]: https://github.com/opencontainers/tob/blame/master/README.md#L21

proposals/distribution.md
yuwaMSFT2: @vsoch no the metadata of the manifest is not what I meant for tenant.
It's more to make the repository hierarchical. Like on dockerhub you can have organization, then under the organization you can have repository. Current uri would be /v2/orgname/reponame/, and the registry treat the whole path component "/orgname/reponame" as name parameter in the route matching. It will be good if we can separate it.

vsoch: @mikebrow  this looks good to me!
@yuwaMSFT2 this is badly needed, and I strongly +1. Right now we clump the entire thing that comes before the digest (after the @) and tag (after the :) and the image name (e.g., ubuntu) as the namespace. But given registry urls and the potential for local registries to decide to use a custom namespace, parsing this string has been [more than challenging](https://github.com/singularityware/singularity/blob/master/libexec/python/shell.py#L82)). Being able to separate these two things? :pray: !!

proposals/distribution.md
samuelkarp: >> The specification has been out and in the open since its inception and we have yet to receive any sizable improvement proposals.
>
> The reason for this may be unrelated to whether people want to make changes, or have worthwhile improvements.
> [...]
> I guess what I'm saying is "let's not close the door on any improvement discussions", especially since this is now going to be an OCI spec and not a Docker one anymore.

To add to what @cyphar stated, I think that the pivot from a Docker spec to an OCI spec changes both the set and focus of the contributors; since the spec is no longer being driven by the Docker product's needs it can attract use-cases that would not have made sense for the Docker product.

@mikebrow That language looks fairly reasonable to me.

proposals/distribution.md
samuelkarp: @yuwaMSFT2 Unless I'm reading the current spec wrong, authentication is [currently included](https://github.com/docker/distribution/tree/master/docs/spec/auth).

proposals/distribution.md
stevvooe: @yuwaMSFT2 A lot of the "extensions" you've referenced are really about content management, rather than content distribution. This API has almost always been about content distribution. Adding these extensions to the specification provides little room for vendors to implement these kinds of functionality in a way that fits their platform.

> Right now we clump the entire thing that comes before the digest (after the @) and tag (after the :) and the image name (e.g., ubuntu) as the namespace.

@vsoch None of this is a part of this specification. Even the hub behavior @yuwaMSFT2 was referencing is the hubs implementation. This specification allows arbitrary paths. Other things, like `latest`, are completely up to the client implementation and the registry doesn't care and this specification doesn't even mention.

Let's make sure this exercise is critiquing the specification and not Docker's implementation of it.

proposals/distribution.md
stevvooe: > Is authentication in the scope of the image distribution?

The authentication for the registry protocol is just http-based authentication. There is a token authentication specification, which may be in scope, but there a lot of ways to actually implement it that may be context-specific.

> Image signing would be very useful for Singularity containers.

This is not in scope. All resources are content-addressable and can be signed in external systems. Early versions of the specification and implementation had integrated signing, but there were a lot of problems with it. In practice, the enforcement needs to be in the hands of the client.

This doesn't mean the registry couldn't be used to store signature blobs, but that would require more thought. In practice, we've found the best approach is to decouple this.

samuelkarp:
AWS is happy to support the adoption of a specification for distribution and I'm happy to serve as a maintainer.

Since the Docker registry protocol has become a de facto standard across implementations of container registries, it does make sense to start there.  However, I think it's important to recognize that there are other OCI users than Docker and that this specification should evolve over time to serve them.

proposals/distribution.md
vsoch: @yuwaMSFT2 agreed about content management vs. distribution, I've never distinguished those two before.

In that the uri is serving as an entry point into a registry, even if it doesn't conform to a specific namespace, tag, arguably I should be able to (knowing the API conforms to the registry) programatically predict the base uri to be calling based on this alone. It could be as general as a general flow of regular expressions to follow, or as specific as a discrete set of different kinds (e.g., abstract uri vs. an arbitrary path). It would be a stronger specification to have some level of predictibility here, otherwise I still need to write a custom thing (after checking!) for each registry endpoint to even interact with most endpoints.

proposals/distribution.md
vsoch: For the token authentication - the fact that it is challenging suggests it's even more important to figure out. I would guess that more rather than fewer would want to use token auth, and perhaps the specification can have a default and then fall back cases for a client to implement that are guaranteed to work for most. Without that, we are in a situation of needing a custom implementation for each one, and that defeats the purpose of having a standard. Maybe we could try, and see how far we get? We can always fall back to sticking with just the basic.

And agreed about the signing, given that it's likely very different! It would be good to come back to this at some future point (and as it's more commonly done and discussed).

wking: > There is a token authentication specification, which may be in scope, but there a lot of ways to actually implement it that may be context-specific.

I'm in favor of moving over enough of the auth spec to allow clients to authenticate with Docker's `Bearer` approach without needing to leave the new project's specs.  I've filed #37 against this PR with the changes I think we need for that and more detailed motivation.

proposals/distribution.md
stevvooe: > agreed about content management vs. distribution, I've never distinguished those two before.

The two systems have vastly different requirements. For example, even though the Hub looks like its the same system, they are logically separate between the registry and the hub ui, in practice. This prevents the registry deployment from getting complicated. All of the UI and management of images gets implemented in a separate system, allowing them to grow in features and functionality.

I'm still not quite understanding what you're talking about regarding URIs: the format that you're talking about parsing has nothing to do with this specification. Different systems may implement completely different behaviors and formats, depending on their opinion.

proposals/distribution.md
stevvooe: The issue with this approach is that each auth provider will then have to issue docker-style tokens to interact with the registry. With a more open approach, each registry can choose which providers are integrated. There is also the issue of the access control model: the model used in docker tokens is specific to the way docker implements the registry. Other providers may want to choose to have different token models.

From the perspective of the client, most of this is opaque. They pull an image and authorize the pull through whatever flow makes sense for the context and provide an opaque value for an http header.

proposals/distribution.md
stevvooe: @vsoch @yuwaMSFT2 I expanded a bit on the divide between content distribution and management in https://github.com/opencontainers/tob/pull/37#issuecomment-360908827. I'll repost here, as there may have been some confusion.

This API had everything in it required to integrate with a container runtime (ie. pull an image, figure out what tags are in a repo, etc.) and APIs required to integrate with content management systems. This includes listing repositories and tags. There is also a notification API that enables in the registry implementation, but isn't a part of this specification (there may be an argument about pulling this in).

dmcgowan:
> However, I think it's important to recognize that there are other OCI users than Docker and that this specification should evolve over time to serve them.

I think we are all OK with evolving and making improvements over time. It is going to be a critical role of the maintainers to realize the impact that any change to this specification can bring. Having spent the last 3 years implementing both sides of this specification, I have seen how difficult it can be to both get clients supporting new features and getting registries to fully implement the specification. Inevitably what happens is clients target a subset of registries and registries target a subset of clients. My hope is that we can first address this by moving to OCI, giving the specification the authority and visibility it needs. After that, we can make extensions **which have been proven out in the wild**, and as a community address the variability of implementations and tendency to implement the lowest common denominator.

vsoch:
@dmcgowan I really like this strategy! And +1+1 for the mental visual of specifications galloping around in the wild! :horse_racing:

runcom:
I would definitely help this out!

proposals/distribution.md
```diff
@@ -0,0 +1,66 @@
+# Abstract
+
+The Docker registry protocol has become the defacto standard across the container registry world ([https://github.com/docker/distribution/blob/master/docs/spec/api.md](https://github.com/docker/distribution/blob/master/docs/spec/api.md)).
+
+In the OCI, having a solid, common distribution specification with conformance testing will ensure long lasting security and interoperability throughout the container ecosystem.
+
```
mikebrow: + This proposal also provides the container ecosystem with a means to discuss and schedule extensions to the distribution specification.

proposals/distribution.md
```diff
@@ -0,0 +1,66 @@
...
+Both of the proposed projects would continue to use existing channels in use by the OCI developer community for communication including:
+
+* GitHub for issues and pull requests
+* The dev@opencontainers.org email list
+* The monthly OCI developer community conference call
+* The #OpenContainers freenode IRC channel
+
+## Versioning / Roadmap
+
+The API spec is currently considered v2 and we will start the specification at v2.0. Fewer places to change and compare, and it would keep with it being a lateral move.
```
AkihiroSuda: I'm not sure we should start with `v2.0`, as the latest `Docker-Distribution-API-Version` header is set to `registry/2.0` (https://docs.docker.com/registry/spec/api/#introduction)

How about starting with `V2 Release 1.0"?

AkihiroSuda: Or just `v2.1`

wking: > Or just v2.1

Somewhat complicating this approach is that the fact that Docker's v2 API has been receiving [lettered minor/patch bumps][1].  Presumably those will all be considered part of the 2.0.x series?

And if the goal is to start using SemVer (which sounds like a useful goal to me), we'll need to specify that and wait for some deprecation period before cutting v2.1, or we'll have all the 2.0 clients choke and die when the [version-check endpoint][2] sets `Docker-Distribution-API-Version` to `registry/2.1`.  We want them to all be able to say “I'm 2.0 compliant, so a 2.1 registry is fine”.

[1]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#changes
[2]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#api-version-check

duglin:
There's been lots of good feedback and suggested edits - would it be possible to get an updated PR with those changes for people to review? I get the sense that we may be getting pretty close to something people can live with.

proposals/distribution.md
stevvooe: @wking In general, since the subversions are backwards compatible, we do not expose the subversions. In general, we should move away from path-level versioning and favor type-level versioning.

proposals/distribution.md
```diff
@@ -0,0 +1,66 @@
...
+
+In the past when the topic of having an OCI specification around the distribution of container images was discussed, it was deferred as "let’s get the image format defined, meanwhile the industry will settle on a distribution standard". Fast forward, OCI image format is out and adopted, and the Registry v2 is the defacto standard. There is and will be use-cases for alternate methods and the future will likely hold creative ways to push, fetch and share container images, but right now this promotion serves to acknowledge by the OCI the current industry standard of distributing container images.
+
+There is polish that is needed e.g. broken links to storage-driver docs, as well as making sections more generic regarding the OCI descriptors and media-types, but on the whole this is a lateral move.
+
+## Initial Maintainers
+
+* Stephen Day <stephen.day@docker.com> (@stevvooe)
+* Vincent Batts <vbatts@redhat.com> (@vbatts)
+* Derek McGowan <derek.mcgowan@docker.com> (@dmcgowan)
```
wking: With this proposal referencing project-template for the project's initial rules, we'll need a way to deal with project-template's [Chief Maintainer][1] role.  Possibilities include:

* Having the TOB select a Chief Maintainer here before putting this proposal up for a vote.
* Landing opencontainers/project-template#20 or some other project-template work that removes the Chief Maintainer role from the project docs before putting this proposal up for a vote.

My recommendation would be to remove the role (slightly more discussion [here][2] and [here][3]), but with that PR has been dangling for 1.4 years, so I don't know how likely it is to land in the next month.  Perhaps parallel work can land the Chief Maintainer removal, or opencontainers/project-template#41 may end up with the TOB in charge of project-template.  With project-template (currently) outside of direct TOB control, the simplest approach would be to appoint a Chief Maintainer here, as that entirely within the TOB's control.

[1]: https://github.com/opencontainers/project-template/blame/651c0ada9d70fed6036cf3e2403bf2ef0fb4df5d/MAINTAINERS_GUIDE.md#L78
[2]: https://github.com/opencontainers/project-template/pull/20#discussion_r95453349
[3]: https://github.com/opencontainers/runtime-spec/pull/420#discussion_r61972504

proposals/distribution.md
vbatts: While there is plenty of discussion here, not such an overhaul needed as to change increase from monthly. I am sure we'll adjust the meeting schedule as needed

proposals/distribution.md
vbatts: @akihirosuda are you saying for that header, or just tagged releases or both?

caniszczyk:
RFC for @opencontainers/tob comment and wider community

We will let this bake for a couple of weeks, I will then ask the @opencontainers/tob for a vote

wking:
On Thu, Mar 01, 2018 at 10:17:07PM +0000, Chris Aniszczyk wrote:
> We will let this bake for a couple of weeks, I will then ask the
> @opencontainers/tob for a vote

There are still some open issues that need TOB input before this goes
to a vote:

* Removing the “additional maintainers to consider” list, possibly
  moving some/all of those entries into the maintainer list [1].
* Selecting (or updating the rules to remove) a Chief Maintainer [2].

I'm pretty sure those are both problems with the proposal (as of
f92fdb384c0d), but it's really up to the TOB to come to a consensus on
their desired solution.

[1]: https://github.com/opencontainers/tob/pull/35#discussion_r163291001
[2]: https://github.com/opencontainers/tob/pull/35#discussion_r165471840

vsoch:
As one of the additional to consider I'm not sure the extent to which I should contribute, so I'll wait to hear more, and listen / watch (:bug: on the :framed_picture:) status for now!

vbatts:
i made a pass over the proposal and will take another pass now.

vbatts:
nice to have your feedback @vsoch !

vbatts:
(and err'body else too :stuck_out_tongue: )

proposals/distribution.md
```diff
@@ -142,8 +142,12 @@ The following entries should remain in the [scope table][scope] but not be addre
 * Allows listing of manifests: docker/distribution#2199
 
 [api.md]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md
+[catalog]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#catalog
 [get-manifest]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#pulling-an-image-manifest
 [iana-auth]: http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
+[image-spec]: https://github.com/opencontainers/image-spec/
```
vsoch: quick question - some of these links are to versioned specs (the ones with the blob and commit) and others are to specific tags, and then others to the blame view. Is this intentional?

wking: > some of these links are to versioned specs (the ones with the blob and commit) and others are to specific tags, and then others to the blame view. Is this intentional?

Yeah.  This spec link floats, so distribution can add more endpoints if the imahe-spec gains more object types.  The Docker links have an explicit commit, because they haven't tagged a release since the last `api.md` comit (I think).  Otherwise I prefer pinning by tag.  Blames are for when I need a specific anchor that the rendered Markdown doesn't provide.

vsoch: ok cool! Just wanted to double check, makes perfect sense.

caniszczyk:
Any other comments from the @opencontainers/tob ?

I'll call for a formal vote next Monday unless I hear any strong objections.

vsoch:
As a slightly comical, but actual experience, I'd like to add a general over-arching goal to develop the spec in such a way so... https://vsoch.github.io/2018/deepvariant/ <-- this doesn't happen :) *insert container API monster here*

So very much looking forward to seeing this progress! :confetti_ball:

estesp:
@caniszczyk I think this is effectively ready for a vote. The editor in me has some wording/grammatical cleanup I would like to do, but I'm not sure how valuable it would be other than readability in a few spots. Would you prefer a brief PR with some edits?

wking:
> The scope table changes are presented in a slightly confusing way, being tucked into a Q&A, but they seem alright.

I agree.  I think most TOB proposals will be about scope table changes, so an explicit section on them makes sense to me.  But I added them to the Q&A to match [the][1] [existing][2] [precedent][3].

[1]: https://github.com/opencontainers/tob/blame/3cfd0452a86bd219c33f444bd766f338a6aeb883/proposals/tools.md#L84
[2]: https://github.com/opencontainers/tob/blame/3cfd0452a86bd219c33f444bd766f338a6aeb883/proposals/selinux.md#L44
[3]: https://github.com/opencontainers/tob/blame/3cfd0452a86bd219c33f444bd766f338a6aeb883/proposals/digest.md#L45

wking:
@vbatts, as a TOB member, do you have any comments on the [open issues][1].  @crosbymichael (another TOB member) [seems to agree][2] that something needs to be done about the “Additional Maintainers to consider” list before a TOB vote (although I'm not clear on what his preferred solution is).

[1]: https://github.com/opencontainers/tob/pull/35#issuecomment-369752168
[2]: https://github.com/opencontainers/tob/pull/35#discussion_r163389472

jonjohnsonjr:
@vsoch deeply sorry for your frustration. If it helps, I had to implement this for GCR on the other side of that API and found it similarly painful. The Accept header juggling has been a huge source of headaches for us... as far as I can tell, the [backwards compatibility bits](https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility) are actually impossible to implement in a way that is externally consistent to clients. :/

Please don't hesistate to reach out to us in the future -- contact info is at the bottom of [this page](https://cloud.google.com/container-registry/docs/support).

FWIW you can fetch the image config with this:

```
curl -L https://gcr.io/v2/deepvariant-docker/deepvariant/blobs/sha256:c0acf3d54dce5eabf6ae422593a3d266e1d5d61129d53f32ec943d133b395a6c
```

The image configuration is uploaded as a blob and served through the **blobs** path, so it's not recognized via the **manifests** path. I believe this is consistent with the registry API.

Hoping we can improve on the distribution API going forward. :+1:

vsoch:
oh thank you @jonjohnsonjr ! It wasn't actually hugely frustrating (I hope it doesn't come across that way) I was on cloud nine pretending to be an investigator to solve the mystery! It was a really good learning experience to reflect on after.... it literally went from ERROR message *12 hours later... what just happened?!* haha.

I didn't know about the support! Does it take some kind of HPC / ticket model? I will try the call again - I *almost* guessed it because I was trying the shasum with the `Docker-Content-Digest` header and didn't realize that the config manifest was in fact a blob.  Onward to good things!! :+1:

stevvooe:
@vsoch @jonjohnsonjr I am not sure I understand the confusion. I understand that things may not seem logically constructed, but there is generally a good reason behind why things are the way the are. If you have specific questions, please reach out to me out of band. We are here to help!

For the most part, you need to include the Accept header on the request to get the right response type when fetching by tag. Most of the behavior you're complaining about are due to the realities of content addressable content and support for schema1 that didn't fully adopt the descriptor model. One can't just rewrite stuff inline for the client because it can no longer verify the digests. schema1 made the mistake of including too many layers of the abstraction in a single data structure and most of the work over the last several years has been about bringing in schema2 which helps with that separation.

As far as `Docker-Content-Digest` is concerned, that is a response-only header that provides secondary verification of fetched resources. This is used in the case where a fetch is made by tag is just provides extra verification between the resource sent and the request headers, preventing backend tampering. It is not required to verify this.

To understand the relationships between the components, you can refer to this:

OCI Components Relationship

The layout of docker components is roughly similar, if not slightly complicated by the original schema1. For this specification, schema1 should be considered irrelevant. If you want more details, @lizrice's blog covers some of the details of schema1 in [her blog](https://medium.com/microscaling-systems/spot-the-docker-difference-9f99adcc4aaf).

In general, the specification could use some clarification on mediatypes. If we phased out the manifests endpoint, in favor of a tags API, we could actually get rid of the main problems with having to provide Accept header. Manifest lists are actually a much better way of dispatching alternatives securely. It lets the client pick, rather than having the registry guess a bunch of things about what the client might want if it is an old version.

In the future, we should really phase out the manifests endpoint and require Content-Type headers for blobs to allow proper GC marking. This will enforce further decoupling from the data types (ie the manifest endpoint is way to coupled to image format). Content can then be generally accessed by first class tags (see the PRs that should be a part of this proposal I referenced earlier).

In general, the further we separate the pure content addressability aspects of the API from the named pointers into that identification space, the better this specification will get.

__TL; DR__ `schema1` was a huge problem.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+
+* “Specifying authentication and authorization schemes”.
+    Docker's current registry uses an [extension][token] of the [`Bearer`][rfc6750] [auth scheme][rfc7235-s2.1].
+    Work on specifying Docker's scheme will continue independently, and is orthogonal to the registry API.
+
+    * What: Specifying authentication and authorization schemes
+    * In/Out/Future: Out of scope
+    * Status: N/A
+    * Description: Defining protocols for authenticating and authorizing distribution access.
+    * Why: As an HTTP-based protocol, clients and servers can negotiate authentication via HTTP's [challenge-response authentication framework][rfc7235-s2.1].
```
stevvooe: This should be specified in such a way that the bearer can be opaque to the registry implementation.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+        There is no need for the distribution specification to choose a particular authentication scheme, because clients receiving 401 and 407 responses can use IANA's [HTTP Authentication Scheme Registry][iana-auth] to look up referenced schemes and take appropriate action.
```
stevvooe: 403, as well. Although, this is too technical for the proposal.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+
+* “Creating a reference spec for optional DNS based naming and discovery”.
```
stevvooe: This seems completely irrelevant.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+    * Description: Define a protocol for resolving an image name to retrieval information.
+        When we address this, we will also allow for alternative name-to-image discovery protocols in parallel with the OCI-specified protocol.
+    * Why: It is reasonable to provide a standardized way to use DNS based distribution in conjunction with OCI without requiring its use.
+        There are many good use cases for DNS based distribution, but not all use cases support this.
+        Furthermore, encoding the location of a bundle into the bundle can cause issues with downloads from alternate locations other than the origin specified in the name.
+
+* “Specifying a distribution method”.
+    This entry replaces part of the previous “Creating Reference spec for optional DNS based naming & distribution” and “Standardizing on a particular Distribution method” entries.
+
+    Retrieving image indexes covers the current “tag listing” (e.g. “what named manifests are in `library/busybox`?”), because tags are entries in the image format's [`manifests` array][manifests].
```
stevvooe: "because tags are entries in the image format's [`manifests` array][manifests]."

This statement is completely false.

stevvooe: This is only true for image layout and has nothing to do with images distributed over this API.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+    Other tag-listing endpoints needed for backwards-compatibility are therefore in scope as well.
+
+    Grouping image indexes in repositories is considered part of distribution policy or content management, which are out of scope for this entry's per-image action.
+    For example, “what images are under `library/`?” is out of scope for this project.
```
stevvooe: From the perspective of this specification, there is no relationship between `library` and `library/foo`. This allows one to arbitrarily map the registry namespace to different schemas. In this specification, they are just treated as a name.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+    * Description: Define a protocol for creating, retrieving, updating, and deleting objects defined in the [image specification][image-spec].
+        Listing repositories (like [`/v2/_catalog`][catalog]) is a multi-[image-index][] action, which is out of scope for this entry.
+        Managing groups of image indexes requires multi-[image-index][] actions, which are out of scope for this entry.
+        Listing image indexes within a group is a multi-[image-index][] action, which is out of scope for this entry.
+    * Why: This specification will provide one (of possibly many) distribution specifications.
+        Alternative distribution specifications may be developed for uses cases not covered by this specification, but defining them is currently out of scope for the OCI.
+
+* “Retrieving image content by its content-addressable hash”.
+    Docker's registry API already provides [endpoints for fetching manifest objects by digest][get-manifest].
+    Docker's registry API does not currently provide endpoints for fetching [image-index][] objects by digest, but this is the project where that will happen.
```
stevvooe: Missing link here.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+
+* Simplifies tag listing: docker/distribution#2169
+* Allows listing of manifests: docker/distribution#2199
+
+[api.md]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md
+[catalog]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#catalog
+[get-manifest]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#pulling-an-image-manifest
+[iana-auth]: http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
+[image-spec]: https://github.com/opencontainers/image-spec/
+[image-index]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md
```
stevvooe: Any reason for the blame links?

vsoch: heyo! I think this was explained a bit up here --> https://github.com/opencontainers/tob/pull/35#discussion_r171951003 (I had the same question!)

proposals/distribution.md
vsoch: I like this approach as well - having developed tools that separated some prefix / collection of a namespace (e.g., library) from the rest (e.g., "foo") led to much error, when really we can just call the whole thing a group (name or namespace?) It's a bit like the "file hierarchy" in Google Storage - it just looks like folders / files but it's really just an index with slashes in it. So I agree -
 instead of having to classify that "library" is a different thing from "library/foo" I think thus it's better to have them treated as a name, and then tools that want to mimic a filesystem (or search, etc.) can do things akin to name.startswith('library') etc.

proposals/distribution.md
dmcgowan: By that explanation this link does not need to be a blame. The other blame link to the same doc has an anchor, but it might be wise to consider not using blame and rather pointing to the parent section so the link is navigable to the sub sections.

jonjohnsonjr:
@stevvooe thanks for the link to Liz's blog post, that's a great resource.

I think we're in violent agreement re: the problems with schema1 :+1:

proposals/distribution.md
wking: > By that explanation this link does not need to be a blame.

Good point.  I've filed #47 against this PR to fix this link.

> The other blame link to the same doc has an anchor, but it might be wise to consider not using blame and rather pointing to the parent section so the link is navigable to the sub sections.

I prefer linking [directly to `manifests`][1] instead of linking to [the whole section][2] (~85 lines of Markdown), but if @caniszczyk and/or the TOB prefer section links I'll survive.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md#L23
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.1/image-index.md#image-index-property-descriptions

proposals/distribution.md
wking: > Missing link here.

This is the “[implicit link name shortcut][1]”, and GitHub [renders it fine][2].

[1]: https://daringfireball.net/projects/markdown/syntax#link
[2]: https://github.com/opencontainers/tob/blob/f92fdb384c0dba03574906800d52038bf661bada/proposals/distribution.md#frequently-asked-questions-faq

proposals/distribution.md
wking: > This should be specified in such a way that the bearer can be opaque to the registry implementation.

Can you file a PR against this branch or suggest alternative wording?  I think making negotiated authentication out of scope (as I've tried to do with this scope entry) is even stronger than “opaque to the registry implementation”.  More on this specific issue in #37, where I initially tried to make the `Bearer` auth part of the scope, but pivoted to the wording here based on @dmcgowan's [comment][1].

[1]: https://github.com/opencontainers/tob/pull/37#issuecomment-360923589

proposals/distribution.md
wking: > 403, as well.

I'm fine with it being added.

> Although, this is too technical for the proposal.

*shrug*.  I think the line above (linking 7235) is sufficient, but this line is more helpful detail for implementation authors.  I'm fine either dropping this line or adding the 403 here and keeping the line.

proposals/distribution.md
wking: > This seems completely irrelevant.

I'm saying that it's out of scope for this project and in scope for future OCI work.  Do you disagree with either of those?  If not, I think we need something in the scope table around discovery/naming to replace the naming portion of the current “Creating Reference spec for optional DNS based naming & distribution” entry (as I say in [the next line][1]).

[1]: https://github.com/opencontainers/tob/pull/35/files#diff-748aea9c83a752af811a80671d8803d2R92

proposals/distribution.md
wking: > "because tags are entries in the image format…"
>
> This statement is completely false.

Would you accept:

```markdown
… because tags can be represented as entries in the image format's [manifests array][manifests].
```

?

proposals/distribution.md
wking: > From the perspective of this specification, there is no relationship between `library` and `library/foo`. This allows one to arbitrarily map the registry namespace to different schemas. In this specification, they are just treated as a name.

That's fine with me.  The point I'm trying to make around here is that if `library/foo` is representable as an image-index, then operations on higher-level objects like `library/` are out of scope for this PR.  There was a lot of wording debate on this in #44 which was eventually resolved in #46, but if you have different wording preferences for that idea I'm open to changes.

proposals/distribution.md
dmcgowan: > This should be specified in such a way that the bearer can be opaque to the registry implementation.

The mentioned rfc is broad enough to cover this case. This section is just saying the registry can return a response requesting a specific authentication type (could be bearer, basic, or whatever). The comment below this could probably be more generalized or removed completely as the rfc already describes this behavior in detail.

proposals/distribution.md
dmcgowan: My vote is to drop it, the previous line mentions negotiation, no need to summarize how the negotiation works.

wking: > I'm fine either dropping this line…

Between this thread and @dmcgowan's [comment above][1], I think there's a consensus around dropping this line.  I've filed #48 to do that.

[1]: https://github.com/opencontainers/tob/pull/35#discussion_r172702752

stevvooe:
> violent agreement

The best kind of agreement! :D

proposals/distribution.md
stevvooe: @wking These are really technical considerations that I don't think belong in the proposal. The only takeaway here is that `_` start is reserved in the repo names to avoid overlap with endpoints like `/v2/_catalog` and to allow space for in-band backend mappings.

proposals/distribution.md
stevvooe: No, because these are not true statements.

proposals/distribution.md
```diff
@@ -0,0 +1,156 @@
...
+    Other tag-listing endpoints needed for backwards-compatibility are therefore in scope as well.
+
+    Grouping image indexes in repositories is considered part of distribution policy or content management, which are out of scope for this entry's per-image action.
+    For example, “what images are under `library/`?” is out of scope for this project.
+
+    * What: Specifying a distribution method
+    * In/Out/Future: In scope
+    * Status: In progress (see opencontainers/distribution-spec)
+    * Description: Define a protocol for creating, retrieving, updating, and deleting objects defined in the [image specification][image-spec].
+        Listing repositories (like [`/v2/_catalog`][catalog]) is a multi-[image-index][] action, which is out of scope for this entry.
```
stevvooe: Again, there is not relation to "multi-image-index" and the catalog endpoint. Please avoid making incorrect statements.

proposals/distribution.md
stevvooe: Let me clarify: repository naming is completely in scope here. The API isn't very useful without it. @vsoch's understanding is correct. How repository names are mapped to image names is up to the implementation or defined by convention.

proposals/distribution.md
wking: > Again, there is not relation to "multi-image-index" and the catalog endpoint. Please avoid making incorrect statements.

How would you prefer to phrase it?  Do you agree that `/v2/_catalog` is a registry-level operation?  Do you agree that it may return multiple repositories?  Do you agree that repositories may contain multiple images?

proposals/distribution.md
wking: > No, because these are not true statements.

[Image-spec says][1]:

> A common use case of descriptors with a "org.opencontainers.image.ref.name" annotation is representing a "tag" for a container image.

Can you go into more detail about how using `manifests` with `org.opencontainers.image.ref.name` annotations is not capable of representing image tags?

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-layout.md#L154

proposals/distribution.md
wking: > These are really technical considerations that I don't think belong in the proposal.

I want to set clear enough scope boundaries that operations like [the current `/v2/_catalog`][1] or the API form of [user's image-index listing][2], while keeping the API form of [tags within an image-index][3] in scope.  Since there's already a `/v2/_catalog` in Docker's `api.md`, I think we need to get technical enough to make that distinction.

> The only takeaway here is that `_` start is reserved in the repo names to avoid overlap with endpoints like `/v2/_catalog` and to allow space for in-band backend mappings.

I'm fine with the OCI distribution API to reserve a namespace like `/v2/_.*` for backend extentions.  I don't think that conflicts with leaving `/v2/_catalog` as a docker/distribution-specific extension.

> Let me clarify: repository naming is completely in scope here. The API isn't very useful without it.

That's fine with me.  What does it have to do with making “[what images are under `library/`?][2]” out of scope for the distribution API?

[1]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#catalog
[2]: https://hub.docker.com/u/library/
[3]: https://hub.docker.com/r/library/alpine/tags/

proposals/distribution.md
stevvooe: I have no clue what is being said here. It literally makes no sense. The specification makes no reference to a "multi-image-index" and we don't have anything called that in the image-spec. It is just made up.

proposals/distribution.md
stevvooe: The thing you linked in the hub is not, at all, an implementation of this specification. The catalog endpoint is just a raw listing of repositories. It's out of scope because it is a management function of the registry and not required for runtimes to pull an image.

proposals/distribution.md
stevvooe: This section is so confusing that its impossible to understand what is intended here. Above it says "FAQ", then it makes a bunch of statements with quotes. What does this effort have to do with naming and discovery?

proposals/distribution.md
stevvooe: Yes, the addition of `ref.name` was a poorly thought addition and doesn't have a relationship here. They are separate things. Yes, they could represent image tags, but the tagging in distribution spec is correctly unique (ie points at a manifest list), whereas `ref.name` tries to embed something. As I've said about a million times, `ref.name` is only valid on the layout `index.json`. If we want to use here as a future change, the maintainers can make that decision. Either way, this statement is wholly confusing.

caniszczyk:
On today's OCI dev call we came to consensus to call the project vote on Monday March 12th for the @opencontainers/tob to make a decision. I'll work with @crosbymichael on the voting procedure

http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2018/opencontainers.2018-03-07-22.00.log.html

caniszczyk:
I kicked off the official project creation vote here for the OCI TOB to approve:

https://groups.google.com/a/opencontainers.org/forum/#!topic/tob/dU6uXM__ilU

cc: @opencontainers/tob

proposals/distribution.md
```diff
@@ -0,0 +1,154 @@
...
+        Using hashing for name is an acceptable addition as it does not encode any centralized namespace.
+
+The following entries should remain in the [scope table][scope] but not be addressed by this project:
+
+* “Specifying way to attach signatures”.
+    We don't need to address this as part of distribution, because all resources are content-addressable and can be signed in external systems.
+
+## Related GitHub Issues
+
+* Simplifies tag listing: docker/distribution#2169
```
stevvooe: Let's add a statement here about what these PRs mean. Are these additions required by the proposal? They have been vetted and implemented, so I would think they are a good candidate.

proposals/distribution.md
jzelinskie: Is this possibly confusion caused by a misunderstanding of ManifestLists?

proposals/distribution.md
jzelinskie: FWIW, Docker v2-2 is the most recently protocol -- I hope everyone here knows that and has just been using 2.1 as an example.

proposals/distribution.md
wking: > As I've said about a million times, `ref.name` is only valid on the layout `index.json`.

In that case, I don't see why tags belong in the distribution spec at all.  If [manifests][1] are the highest-level object that is represented in the API, then why bother with tag-listing at all?  Just require consumers to ask for exactly the manifest they need with something like:

    $ curl -H 'Accept: application/vnd.oci.image.manifest.v1+json' /v2/library/alpine/3.7

if, instead, you want to continue to provide [an endpoint that lists multiple tags][2], is there really a benefit to returning a list of strings instead of a [list of descriptors][3]?


[1]: https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md
[2]: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#listing-image-tags
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md#L23

dmcgowan: > In that case, I don't see why tags belong in the distribution spec at all

That is something we could consider for enhancing the protocol. Right now the API can mostly be used with tags but manifests cannot be uploaded without a tag. It is best to think of manifests as the highest level object and tags as part of the API for usability. Tags separated from the API is sometimes useful for security, this is what notary was created to solve.

As for this API, can you please remove language linking tags with the manifests and indexes, that is confusing as there is no relationship within the API. You can keep language about tag listing endpoints as being in scope, although as mentioned, there are other ways to get lists of tags from external indexes.

proposals/distribution.md
stevvooe: These are unrelated concepts.

jzelinskie: Rereading the sentence, it sounds there's an assumption that the registry implements a 'multi-image index' (e.g. an indexed repository table in a relational database) used to serve the catalog endpoint. This document should avoid making any assumptions about the registry implementation.

proposals/distribution.md
stevvooe: @wking Can you fix the wording here? This should only call out that hierarchies aren't a part of this specification.

proposals/distribution.md
jzelinskie: I think you can forgo the `[]` and just leave it `[image-name]` and it'll continue to work, as well.

proposals/distribution.md
wking: > * Landing opencontainers/project-template#20 or some other project-template work that removes the Chief Maintainer role from the project docs before putting this proposal up for a vote.

This happened, so nothing left to do for this specific point.

proposals/distribution.md
wking: @caniszczyk, were we just dropping this list, and leaving it up to the above-three maintainers to add additional maintainers?

proposals/distribution.md
stevvooe: This has been fixed under #50.

wking:
There's a second run at the vote going on [here][1].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/tob/sJ7_H72M9FM

vbatts:
LGTM

caniszczyk:
The vote PASSED:

https://groups.google.com/a/opencontainers.org/forum/#!topic/tob/sJ7_H72M9FM

PR #38, opened 2018-01-26
README: Link to the distribution proposal

wking:
In case merging a pull-request is the easiest way to address [this][1].  I'm also fine with @caniszczyk making this update directly to his PR, or squashing this commit into his branch, or whatever.

[1]: https://github.com/opencontainers/tob/pull/35#issuecomment-360243769


caniszczyk:
fine with me

PR #47, opened 2018-03-06
distribution: Change from blame to rendered URI for image-index

wking:
We only need blame views when the Markdown does not provide a specific-enough anchor, and this is a link to the whole file.

Fixes the issue raised be @stevvooe [here][1].

[1]: https://github.com/opencontainers/tob/pull/35#discussion_r172689573


dmcgowan:
LGTM

wking:
@caniszczyk, did you want to merge this before the distribution vote hits its stride?

stevvooe:
LGTM

PR #48, opened 2018-03-07
distribution: Remove IANA auth-scheme sentence

wking:
@stevvooe [points out][1] that the response code list was missing 403 and that the line is fairly low-level.  I agree with @dmcgowan [that][2] that the RFC 7235 reference is sufficient to express the idea, and the less we say, the less likely we are to be wrong ;).

[1]: https://github.com/opencontainers/tob/pull/35#discussion_r172687857
[2]: https://github.com/opencontainers/tob/pull/35#discussion_r172702752


dmcgowan:
LGTM

stevvooe:
LGTM

PR #49, opened 2018-03-14
proposals/distribution: remove non-sense about image indexes

stevvooe:
Signed-off-by: Stephen J Day <stephen.day@docker.com>

### DIFF
--- a/proposals/distribution.md
+++ b/proposals/distribution.md
@@ -112,10 +112,8 @@ The following entries should be added to the [scope table][scope]:
     * What: Specifying a distribution method
     * In/Out/Future: In scope
     * Status: In progress (see opencontainers/distribution-spec)
-    * Description: Define a protocol for creating, retrieving, updating, and deleting objects defined in the [image specification][image-spec].
-        Listing repositories (like [`/v2/_catalog`][catalog]) is a multi-[image-index][] action, which is out of scope for this entry.
-        Managing groups of image indexes requires multi-[image-index][] actions, which are out of scope for this entry.
-        Listing image indexes within a group is a multi-[image-index][] action, which is out of scope for this entry.
+    * Description: Define a protocol for creating, retrieving, updating, and deleting content-addressable objects, such as those defined in the [image specification][image-spec].
+		Listing repositories is a registry-wide operation; the implementation of which is out of scope of this document.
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

proposals/distribution.md
```diff
@@ -112,10 +112,8 @@ The following entries should be added to the [scope table][scope]:
     * What: Specifying a distribution method
     * In/Out/Future: In scope
     * Status: In progress (see opencontainers/distribution-spec)
-    * Description: Define a protocol for creating, retrieving, updating, and deleting objects defined in the [image specification][image-spec].
-        Listing repositories (like [`/v2/_catalog`][catalog]) is a multi-[image-index][] action, which is out of scope for this entry.
-        Managing groups of image indexes requires multi-[image-index][] actions, which are out of scope for this entry.
-        Listing image indexes within a group is a multi-[image-index][] action, which is out of scope for this entry.
+    * Description: Define a protocol for creating, retrieving, updating, and deleting content-addressable objects, such as those defined in the [image specification][image-spec].
+		Listing repositories is a registry-wide operation; the implementation of which is out of scope of this document.
```
mikebrow: Nit: this is a first introduction of term "document." Suggest using this "spec."
Also this listing restriction does not make sense here. Can we move it to the Grouping line above on 109 and change the word grouping to a broader term like "managing?"

dmcgowan: The relationship between these two sentences is not straightforward. However I do think this is referring to slightly different functionality. While the "grouping" section is just saying that each repository name is considered independent from the perspective of the API, this is referring to an endpoint which lists all the repository names within a registry. While the group and name hierarchy could be of interest to an end API user, cataloging has normally just be considered an administrative endpoint not exposed to API users.

+1 on document -> specification

mikebrow: Cool..
Maybe.. "An endpoint for listing repository names within a registry, is considered out of scope for this spec and thus also out of scope for this protocol."

dmcgowan: That language sounds clearer, let me update and move it next to tag listing. I don't think these statements belong here.

dmcgowan:
Pulled this commit into #50, please close this PR

crosbymichael:
closing per request with changes in #50

PR #53, opened 2018-12-13
Add CNAB proposal

lachie83:


lachie83:
Associated mailing-list thread - https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/qYM84b0vays

technosophos:
In the Jan 16, 2019 meeting, we briefly discussed this item (though it was not mentioned in the [minutes](http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2019/opencontainers.2019-01-16-22.02.txt)). Someone suggested that there was a proposal to role CNAB into the image specification. I cannot find any reference to a conversation about that.

My preference would be to not try to conflate the two. While there is an affinity between CNAB's format and the image spec, there is no sense in which one cannot write an image without CNAB or cannot write a CNAB without an OCI image. That feels like a natural indicator that they are not part of the same specification.

mikebrow:
See opencontainers.slack.com #opencontainers channel for some links / discussions around this topic.

garethr:
@mikebrow how does one join that Slack? I don't see the Slack channel mentioned on the [org](https://github.com/opencontainers/org) repo and signups on https://opencontainers.slack.com/signup are only for those with a Linux Foundation email address?

caniszczyk:
Use slack.opencontainers.org to request an invite but all the action happens on IRC

technosophos:
Thanks @mikebrow! The discussion during the meeting had only one mention of the CNAB spec, and it was this:

> 23:00:51 <vbatts|work> steveoovoeoeve: there was a proposal for cnab-spec to the tob, that should closed/rolled in

But we don't know what @stevvooe was pointing to when he said there was a proposal to "roll it in", and we're not sure what it would be rolled into.

For a tiny bit of context, there are two distinct discussions that are involving CNAB:

1. _Are there changes to the registry for storing CNAB (and other) artifacts? If so, what?_ That conversation seems near resolution.
2. Ought the CNAB spec become part of OCI specifications?_ This deals with CNAB as a format, and is relatively independent of the discussion of where to store artifacts.

This thread is to discuss option 2.

dmcgowan:
-1

I think this can now be achieved within the scope of #60

Issue #66, opened 2020-02-14
New project: opencontainers/oras


jdolitsky:
## Summary

Adopt the ORAS project located at [deislabs/oras](https://github.com/deislabs/oras).

## Overview

ORAS is a CLI that can publish arbitrary content to an OCI registry, with special features for setting mediatypes on manifest configs and on content.

Note: the manifest mediatype itself is always `application/vnd.oci.image.manifest.v1+json`.

Example - uploading rockets, a brand new type of package:

```
# Create a thing
printf '🚀' > rocket.txt

# Create a manifest config
printf '{"RocketVersion":"v0.1.0"}' > rocket-config.json

# Upload your thing with a custom mediatype
oras push localhost:5000/mystuff/myrocket:v0.1.0 rocket.txt:text/plain \
  --manifest-config rocket-config.json:application/vnd.acme.rocket.config.v1+json
 ```

See manifest created:

```
$ curl -s -H 'Accept: application/vnd.oci.image.manifest.v1+json' \
    http://localhost:5000/v2/mystuff/myrocket/manifests/v0.1.0 | jq
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.acme.rocket.config.v1+json",
    "digest": "sha256:310175f34d2d4d5cba3418be06ddd1ef948147d729516d78318ec7f5c2d83d49",
    "size": 26
  },
  "layers": [
    {
      "mediaType": "text/plain",
      "digest": "sha256:ebbc0b2870eb323f2b6cffa5c493ceef81ae7eb36afc73d4e0367301631daec5",
      "size": 4,
      "annotations": {
        "org.opencontainers.image.title": "rocket.txt"
      }
    }
  ]
}
```

Get that thing:

```
$ curl -s http://localhost:5000/v2/mystuff/myrocket/blobs/sha256:ebbc0b2870eb323f2b6cffa5c493ceef81ae7eb36afc73d4e0367301631daec5
🚀
```

## Additional Usage

ORAS is built primarily on top of Go packages provided by [containerd](github.com/containerd/containerd), but it also imports packages from the [docker/cli](https://github.com/docker/cli), which enables "docker-style" auth login:

```
oras login -u username -p password localhost:5000 -c rocket-creds.json
```

There are also public Go packages available to build on top of ORAS. The following is the equivalent of the rocket example with the CLI above, but in Go:

```go
package main

import (
	"context"
	"fmt"

	"github.com/containerd/containerd/remotes/docker"
[...]

PR #68, opened 2020-02-24
projects: add ORAS proposal

jdolitsky:
Continuation of #66

**Vote Total (Closes 2020-06-19)** (please make sure you also leave an LGTM comment):
- [x] Vincent Batts (@vbatts)
- [ ] Michael Crosby (@crosbymichael)
- [x] Phil Estes (@estesp)
- [ ] Wei Fu (@fuweid)
- [ ] Jon Johnson (@jonjohnsonjr)
- [ ] Samuel Karp (@samuelkarp)
- [x] Steve Lasker (@SteveLasker)
- [ ] Derek McGowan (@dmcgowan)
- [ ] Aleksa Sarai (@cyphar)

/cc @opencontainers/tob

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ https://groups.google.com/a/opencontainers.org/forum/#!forum/tob (tob@opencontai
 * [Image Format Spec](proposals/image-format)
 * [SELinux](proposals/selinux.md)
 * [Tools](proposals/tools.md)
+* [ORAS](proposals/oras.md)
 
 ## Voting
 
--- a/proposals/oras.md
+++ b/proposals/oras.md
@@ -1,0 +1,214 @@
+# OCI ORAS Project Proposal #
+
+## Abstract ##
+
+In order to provide OCI end users a method to publish and retrieve any type of content to/from an OCI registry, as well as a reference implementation for the OCI Distribution Specification, the [ORAS project][oras-project] should be moved under the opencontainers GitHub org.
+
+[oras-project]: https://github.com/deislabs/oras
+
+### ORAS Details ###
+
+ORAS is a CLI that can publish arbitrary content to an OCI registry, with special features for setting mediatypes on manifest configs and on content.
+
+Note: the manifest mediatype itself is always `application/vnd.oci.image.manifest.v1+json`.
+
+Example - uploading rockets, a brand new type of package:
+
+```
+# Create a thing
+printf '🚀' > rocket.txt
+
+# Create a manifest config
+printf '{"RocketVersion":"v0.1.0"}' > rocket-config.json
+
+# Upload your thing with a custom mediatype
+oras push localhost:5000/mystuff/myrocket:v0.1.0 rocket.txt:text/plain \
+  --manifest-config rocket-config.json:application/vnd.acme.rocket.config.v1+json
+ ```
+
+See manifest created:
+
+```
+$ curl -s -H 'Accept: application/vnd.oci.image.manifest.v1+json' \
+    http://localhost:5000/v2/mystuff/myrocket/manifests/v0.1.0 | jq
+{
+  "schemaVersion": 2,
+  "config": {
+    "mediaType": "application/vnd.acme.rocket.config.v1+json",
+    "digest": "sha256:310175f34d2d4d5cba3418be06ddd1ef948147d729516d78318ec7f5c2d83d49",
+    "size": 26
+  },
+  "layers": [
+    {
+      "mediaType": "text/plain",
+      "digest": "sha256:ebbc0b2870eb323f2b6cffa5c493ceef81ae7eb36afc73d4e0367301631daec5",
+      "size": 4,
+      "annotations": {
+        "org.opencontainers.image.title": "rocket.txt"
+      }
+    }
+  ]
+}
+```
+
+Get that thing:
+
+```
+$ curl -s http://localhost:5000/v2/mystuff/myrocket/blobs/sha256:ebbc0b2870eb323f2b6cffa5c493ceef81ae7eb36afc73d4e0367301631daec5
+🚀
+```
+
+#### Additional Usage ####
+
+ORAS is built primarily on top of Go packages provided by [containerd][containerd-project], but it also imports packages from the [docker/cli][dockercli-project], which enables "docker-style" auth login:
+
+```
+oras login -u username -p password localhost:5000 -c rocket-creds.json
+```
+
+There are also public Go packages available to build on top of ORAS. The following is the equivalent of the rocket example with the CLI above, but in Go:
+
+```go
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/containerd/containerd/remotes/docker"
+	"github.com/deislabs/oras/pkg/content"
+	"github.com/deislabs/oras/pkg/oras"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+func main() {
+	ctx := context.Background()
+	resolver := docker.NewResolver(docker.ResolverOptions{})
+	store := content.NewMemoryStore()
+
+	registryRootURL := "localhost:5000"
+	registryNamespace := "mystuff/myrocket"
+
+	rocketVersion := "v0.1.0"
+	rocketFileName := "rocket.txt"
+	rocketMediaType := "text/plain"
+	rocketContent := []byte("🚀")
+	rocketDescriptor := store.Add(rocketFileName, rocketMediaType, rocketContent)
+
+	rocketConfigMediaType := "application/vnd.acme.rocket.config.v1+json"
+	rocketConfigContent := []byte(fmt.Sprintf("{\"RocketVersion\":\"%s\"}", rocketVersion))
+	rocketConfigDescriptor := store.Add("", rocketConfigMediaType, rocketConfigContent)
+
+	ref := fmt.Sprintf("%s/%s:%s", registryRootURL, registryNamespace, rocketVersion)
+	_, err := oras.Push(ctx, resolver, ref, store, []ocispec.Descriptor{rocketDescriptor},
+		oras.WithConfig(rocketConfigDescriptor))
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Pushed to", ref)
+	fmt.Printf("\nTry:\n\ncurl -s -H 'Accept: application/vnd.oci.image.manifest.v1+json' \\\n" +
+		"    %s/v2/%s/manifests/%s | jq\n", registryRootURL, registryNamespace, rocketVersion)
+}
+```
+
+You can see all features in the project [README][oras-readme].
+
+[containerd-project]: https://github.com/containerd/containerd
+[dockercli-project]: https://github.com/docker/cli
+[oras-readme]: https://github.com/deislabs/oras/blob/master/README.md
+
+## Proposal ##
+Change the ownership of the existing ORAS project from deislabs:
+
+  https://github.com/deislabs/oras
+
+And move it inside the `opencontainers` organization:
+
+  https://github.com/opencontainers/oras
+
+The import paths will correspondingly be "github.com/opencontainers/oras" (oras does have some Go API users, but since the project will be renamed -- and GitHub will add a redirect -- there will be no significant downstream impact of the change).
+
+### Initial Maintainers ###
+Initial maintainers of the ORAS project would be:
+
+* Josh Dolitsky <josh@bloodorange.io> (@jdolitsky)
+* Shiwei Zhang <shizh@microsoft.com> (@shizhMSFT)
+* Sajay Antony <sajaya@microsoft.com> (@sajayantony)
+* Steve Lasker <steve.lasker@microsoft.com> (@stevelasker)
+* Jimmy Zelinskie <jimmyzelinskie@gmail.com> (@jzelinskie)
+* Vincent Batts <vbatts@hashbangbash.com> (@vbatts)
+
+### Code of Conduct ###
+This project would incorporate (by reference) the OCI [Code of Conduct][code-of-conduct].
+
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/CODE_OF_CONDUCT.md
+
+### Governance and Releases ###
+This project would incorporate the Governance and Releases processes from the [OCI project template][oci-project-template].
+
+It should be noted that since ORAS is not a specification, it is not bound by the ordinary quorum and voting rules for specification release.
+As such, new versions will be released as regularly as needed without the need for a quorum vote.
+
+Pull requests will require two (2) reviews (signified by "LGTM") from project maintainers.
+Maintainers are not allowed to review a pull request which they authored.
+
+[oci-project-template]: https://github.com/opencontainers/project-template
+
+### Project Communications ###
+The proposed project would continue to use existing channels in use by the OCI developer community for communication including:
+
+ * GitHub for issues and pull requests.
+ * The [`dev@opencontainers.org`][oci-ml] email list.
+ * The weekly OCI developer community conference call.
+ * The `#opencontainers` IRC channel on Freenode.
+ * The [OCI Slack workspace][oci-slack].
+ * The [OCI Matrix Room][oci-matrix].
+
+[oci-ml]: mailto:dev@opencontainers.org
+[oci-slack]: https://opencontainers.slack.com/
+[oci-matrix]: https://matrix.to/#/#opencontainers:matrix.org
+
+## Frequently Asked Questions (FAQ)
+> *Does this proposal change the OCI Charter?*
+
+This proposal does not in any way intend to amend the [OCI Charter][oci-charter].
+
+[oci-charter]: https://github.com/opencontainers/tob/blob/master/CHARTER.md
+
+> *Where does ORAS fit into the OCI suite of projects?*
+
+ORAS is intended to be a *reference implementation of the OCI Distribution Specification*.
+
+As ORAS was designed to handle any type of content, it will serve to exercise the spec
+to make it more independent of details that may have leaked in from image-spec or Docker.
+
+> *Why bless ORAS over other alternative tools?*
+
+ORAS has already been used successfully in the wild as a building block for 
+publishing custom content to an OCI registry.
+
+The following projects are already successfully using ORAS to work with custom artifacts:
+
+- [Helm][helm-usage]
+- [Conftest][conftest-usage]
+- [Singularity][singularity-usage]
+
+[helm-usage]: https://github.com/helm/helm/search?q=oras
+[conftest-usage]: https://github.com/instrumenta/conftest/search?q=oras
+[singularity-usage]: https://github.com/sylabs/singularity/search?q=oras
+
+> *How do we avoid the runc issue with implementation-specific quirks becoming a de-facto standard?*
+
+When developing new features in ORAS (which are within the scope of the OCI Distribution Specification),
+a strong effort will be made to include those features in the upstream specification.
+
+> *Who are the other target users of ORAS?*
+
+Users seeking a common way to store different types of content (not just container runtime images),
+using the OCI Distribution Spec as the baseline API.
+
+> *How do you pronounce ORAS?*
+
+The name ORAS is actually an acronym for "OCI Registry As Storage",
+but when speaking of it, you can say "or-ahs".


proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Abstract ##
+
+The ORAS project (https://github.com/deislabs/oras) is a CLI that can publish arbitrary content to an OCI registry, with special features for setting mediatypes on manifest configs and on content. 
+
+In order to provide OCI end users a method to publish and retrieve any type of content to/from an OCI registry, as well as a reference implementation for the in-progress artifacts spec, ORAS should be moved under the opencontainers GitHub org.
+
+### ORAS Details ###
+
+
```
ArangoGutierrez: Markdown only requires 1 line after titles not 2
```suggestion
```

proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+#### Adoption ####
+
+The following projects are already successfully using ORAS to work with custom artifacts:
+
+- [Helm](https://github.com/helm/helm/search?q=oras)
+- [Conftest](https://github.com/instrumenta/conftest/search?q=oras)
+- [Singularity](https://github.com/sylabs/singularity/search?q=oras)
+
+
```
ArangoGutierrez:
```suggestion
```

proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+}
+```
+
+Get that thing:
+
+```
+$ curl -s http://localhost:5000/v2/mystuff/myrocket/blobs/sha256:ebbc0b2870eb323f2b6cffa5c493ceef81ae7eb36afc73d4e0367301631daec5
+🚀
+```
+
```
ArangoGutierrez:
```suggestion
```

proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+The following projects are already successfully using ORAS to work with custom artifacts:
+
+- [Helm](https://github.com/helm/helm/search?q=oras)
+- [Conftest](https://github.com/instrumenta/conftest/search?q=oras)
+- [Singularity](https://github.com/sylabs/singularity/search?q=oras)
+
+
+
+## Proposal ##
+Change the ownership of the existing umoci project from deislabs:
```
cyphar: `s/umoci/oras/g`. :wink:

proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/CODE_OF_CONDUCT.md
+
+### Governance and Releases ###
+This project would incorporate the Governance and Releases processes from the OCI project template: https://github.com/opencontainers/project-template.
+
+It should be noted that since ORAS is not a specification, it is not bound by the ordinary quorum and voting rules for specification release.
+As such, new versions will be released as regularly as needed without the need for a quorum vote.
+
+Until there are enough additional maintainers, PRs will be merged with only one maintainer LGTM required.
```
cyphar: This doesn't quite appear to match the maintainers list (which has a fair number of people) -- two LGTMs shouldn't be an unreasonable requirement for PRs.

proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+* Jimmy Zelinskie <jimmyzelinskie@gmail.com> (@jzelinskie)
+* Matt Fisher <matt.fisher@microsoft.com> (@bacongobbler)
+
+### Code of Conduct ###
+This project would incorporate (by reference) the OCI [Code of Conduct][code-of-conduct].
+
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/CODE_OF_CONDUCT.md
+
+### Governance and Releases ###
+This project would incorporate the Governance and Releases processes from the OCI project template: https://github.com/opencontainers/project-template.
```
cyphar: Should be a markdown link, not a raw URL.

proposals/oras.md
```diff
@@ -0,0 +1,177 @@
...
+ * The [`dev@opencontainers.org`][oci-ml] email list.
+ * The weekly OCI developer community conference call.
+ * The `#opencontainers` IRC channel on Freenode.
+ * The [OCI Slack workspace][oci-slack].
+ * The [OCI Matrix Room][oci-matrix].
+
+[oci-ml]: mailto:dev@opencontainers.org
+[oci-slack]: https://opencontainers.slack.com/
+[oci-matrix]: https://matrix.to/#/#opencontainers:matrix.org
+
```
cyphar: FAQ section is missing.

proposals/oras.md
ArangoGutierrez: 2  ++

proposals/oras.md
jdolitsky: :)

jdolitsky:
@cyphar @ArangoGutierrez comments addressed! thanks

proposals/oras.md
```diff
@@ -0,0 +1,196 @@
...
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/CODE_OF_CONDUCT.md
+
+### Governance and Releases ###
+This project would incorporate the Governance and Releases processes from the [OCI project template][oci-project-template].
+
+It should be noted that since ORAS is not a specification, it is not bound by the ordinary quorum and voting rules for specification release.
+As such, new versions will be released as regularly as needed without the need for a quorum vote.
+
+PRs will require two (2) LGTMs from project maintainers
+Maintainers are encouraged to not merge their own PRs immediately (preferably waiting at least 24 to 48 hours to allow another maintainer to review it), but this is not a strict rule.
```
cyphar: This line is no longer necessary -- by definition you can't merge your own PR without other people reviewing it if you require two LGTMs.

proposals/oras.md
```diff
@@ -0,0 +1,196 @@
...
+
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/CODE_OF_CONDUCT.md
+
+### Governance and Releases ###
+This project would incorporate the Governance and Releases processes from the [OCI project template][oci-project-template].
+
+It should be noted that since ORAS is not a specification, it is not bound by the ordinary quorum and voting rules for specification release.
+As such, new versions will be released as regularly as needed without the need for a quorum vote.
+
+PRs will require two (2) LGTMs from project maintainers
```
cyphar: Two points:

 1. You should clarify whether maintainers are allowed to "self-LGTM" (as in, are they allowed to LGTM a pull request which they authored)? Technically this is not permitted by the OCI project template, but it's up to each OCI project to decide whether this is a reasonable maintainership model.

 2. This should probably be written as:

```suggestion
Pull requests will require two (2) reviews (signified by "LGTM") from project maintainers.
```

jdolitsky:
@cyphar sorry for delay, comments addressed

proposals/oras.md
```diff
@@ -0,0 +1,196 @@
...
+  https://github.com/deislabs/oras
+
+And move it inside the `opencontainers` organization:
+
+  https://github.com/opencontainers/oras
+
+The import paths will correspondingly be "github.com/opencontainers/oras" (oras does have some Go API users, but since the project will be renamed -- and GitHub will add a redirect -- there will be no significant downstream impact of the change).
+
+### Initial Maintainers ###
+Initial maintainers of the ORAS project would be:
```
vbatts, line 147: A) This may be too many folks from a single employer
B) I could participate here

vbatts:
apart from the maintainer balance, :+1:

proposals/oras.md
SteveLasker: Since you're no longer at RedHat, that would make 4 companies.

SteveLasker: We did reduce Matt Fisher - see pr https://github.com/deislabs/oras/pull/162

cyphar:
@SteveLasker You probably want to squash the PR to a single commit for the proposal.

estesp:
@SteveLasker can you add the same voting markdown from #67 to the top comment so we can align these 2 proposals with the Friday vote deadline?

Thanks!

SteveLasker:
Just getting several cleanup items done, but added the voting to follow the dates committed.

SteveLasker:
> @SteveLasker You probably want to squash the PR to a single commit for the proposal.

They're actually 2, one from @jdolitsky the other from me. These can get squashed on merge, no?

proposals/oras.md
```diff
@@ -0,0 +1,195 @@
...
+ * The [OCI Slack workspace][oci-slack].
+ * The [OCI Matrix Room][oci-matrix].
+
+[oci-ml]: mailto:dev@opencontainers.org
+[oci-slack]: https://opencontainers.slack.com/
+[oci-matrix]: https://matrix.to/#/#opencontainers:matrix.org
+
+## Frequently Asked Questions (FAQ)
+*Q: Does this change the OCI Charter or Scope Table?*
+A: Yes, this will add an additional row to the [OCI Scope Table][oci-scope-table], which does not currently include any mention of generic OCI image tooling.
```
cyphar, line 187: This line needs to be updated now that there isn't an OCI Scope Table anymore (as I did in #67). In addition, ORAS should probably be referred to as an "OCI distribution tool"?

You should also answer some of the new questions I've added to #67 -- which effectively boils down to "what category of project is this (as defined by #76 and the still-draft #86)?". The reason this is important is that it's not quite clear whether this intends to be a Reference Implementation of *something* (whether that be distribution-spec as a client, or artifacts) or a Library, and the scope and usage of the project depends a fair bit on that categorisation. My view is that this appears to be a Reference Implementation (because while it is usable as a Library that's effectively an implementation detail -- umoci is also usable as a library, as is runc).

You also should explain how you plan to avoid the "runc issue" (conventions in a Reference Implementation that are within the scope of the Specification becoming a de-facto standard even though they weren't standardised and thus are impossible to change).

proposals/oras.md
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/containerd/containerd/remotes/docker"
+	"github.com/deislabs/oras/pkg/content"
+	"github.com/deislabs/oras/pkg/oras"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+func main() {
+	ctx := context.Background()
+	resolver := docker.NewResolver(docker.ResolverOptions{})
```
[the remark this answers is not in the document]
jdolitsky: I've just opened https://github.com/deislabs/oras/issues/165 to track this

proposals/oras.md
```diff
@@ -0,0 +1,195 @@
...
+    "digest": "sha256:310175f34d2d4d5cba3418be06ddd1ef948147d729516d78318ec7f5c2d83d49",
+    "size": 26
+  },
+  "layers": [
+    {
+      "mediaType": "text/plain",
+      "digest": "sha256:ebbc0b2870eb323f2b6cffa5c493ceef81ae7eb36afc73d4e0367301631daec5",
+      "size": 4,
+      "annotations": {
+        "org.opencontainers.image.title": "rocket.txt"
```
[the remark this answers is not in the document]
jdolitsky: This is the default. It is basically the way oras knows how to write files to the current directory on pull.

However, this can be disabled by using custom layers/descriptors and `oras.WithNameValidation(nil)` on push.

The annotation itself was take from image-spec, see [here](https://github.com/opencontainers/image-spec/blob/da296dcb1e473a9b4e2d148941d7faa9ac8fea3f/annotations.md).

I wouldn't consider this a part of any spec surrounding artifacts, its just the way oras CLI chooses to store filenames. In the case of Helm, for example, it is entirely unused despite using oras as a library.

jdolitsky:
I've squashed Steve's commit which fixed DCO check

proposals/oras.md
```diff
@@ -0,0 +1,195 @@
...
+
+The [ORAS project][oras-project] is a CLI that can publish arbitrary content to an [OCI Spec Compliant](https://github.com/opencontainers/oci-conformance) registry, with special features for setting mediatypes on manifest configs and on content. 
+
+In order to provide OCI end users a method to publish and retrieve any type of content to/from an OCI registry, as well as a reference implementation for the artifacts project, ORAS should be moved under the opencontainers GitHub org.
+
+[oras-project]: https://github.com/deislabs/oras
+
+### ORAS Details ###
+
+ORAS is a CLI that can publish arbitrary content to an OCI registry, with special features for setting mediatypes on manifest configs and on content.
```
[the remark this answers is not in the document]
jdolitsky: yes, you're right. I have gone ahead and kept it here and simplified the first sentence

proposals/oras.md
jdolitsky: Just updated FAQ section, stating ORAS a reference implementation of distribution spec

jdolitsky:
Update: addressed Sam and Aleksa's comments by updating the introduction and FAQ, also added Vince as initial maintainer.

proposals/oras.md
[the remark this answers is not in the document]
SteveLasker: The annotations are something we've debated. I'm not a big fan of them, and these should likely be something that went into the config for a specific artifact type. That said, it's interesting to see what's possible.

SteveLasker:
> apart from the maintainer balance, LGTM

@vbatts, looks like the squash removed the [x] vote

proposals/oras.md
SteveLasker: We've also said it was a library, to be used by other tools. I suppose the question is whether it can be in both categories.

proposals/oras.md
```diff
@@ -0,0 +1,214 @@
...
+## Frequently Asked Questions (FAQ)
+> *Does this proposal change the OCI Charter?*
+
+This proposal does not in any way intend to amend the [OCI Charter][oci-charter].
+
+[oci-charter]: https://github.com/opencontainers/tob/blob/master/CHARTER.md
+
+> *Where does ORAS fit into the OCI suite of projects?*
+
+ORAS is intended to be a *reference implementation of the OCI Distribution Specification*.
```
samuelkarp, line 181: This is really hard for me to wrap my head around.  As far as I can understand, containerd's `dockerResolver` seems to be the actual OCI distribution client that is being used as part of ORAS.

It seems to me that ORAS is more of a library and tool for enabling artifact-related workflows with an OCI distribution.

proposals/oras.md
```diff
@@ -0,0 +1,214 @@
...
+- [Singularity][singularity-usage]
+
+[helm-usage]: https://github.com/helm/helm/search?q=oras
+[conftest-usage]: https://github.com/instrumenta/conftest/search?q=oras
+[singularity-usage]: https://github.com/sylabs/singularity/search?q=oras
+
+> *How do we avoid the runc issue with implementation-specific quirks becoming a de-facto standard?*
+
+When developing new features in ORAS (which are within the scope of the OCI Distribution Specification),
+a strong effort will be made to include those features in the upstream specification.
```
samuelkarp, line 204: Can ORAS commit to ensuring that non-specified features remain marked as strictly experimental and unsupported until such time as those features are included in the upstream specification?

proposals/oras.md
jdolitsky: Yes, I sort of agree here.. I was going off of Aleksa's suggestion. 

Any issue marking this as a library vs. reference implementation? Does marking it one way or another have any effect? @cyphar @SteveLasker

proposals/oras.md
jdolitsky: @samuelkarp sure. Meaning from a readme/documentation perspective? I'm assuming this in reference to keying off of `config.mediaType` for determining artifact types

samuelkarp: Yes, and discouraging use.  The concern I have here is the same as the concern about the annotations: I'm worried about non-standardized behavior of a reference implementation becoming a de-facto requirement.

proposals/oras.md
cyphar: Yes, there is a distinction. From my point of view, ORAS cannot be argued to be a library because it isn't small enough nor is it merely a tool to assist the implementation of OCI specifications. It is (ostensibly) a (reference) implementation of the distribution-spec that happens to have a bunch of other features. You could argue that the `dockerResolver` of containerd is also a (reference) implementation of the distribution-spec. Not to mention it has a CLI, but I don't want to give the impression that that is the sticking point.

You could argue we need another category for "things that aren't reference implementations" but I think that would dilute the mission of the OCI too far. If you both feel that ORAS could be argued as a library then you should comment on #76 (and eventually #86) to express your disagreement with the current text of both documents.

proposals/oras.md
cyphar: I would go further than just documentation. You should have to pass an actual `--unsupported` or `--experimental` or `--i-understand-this-voids-my-warranty` flag. That is what I intend to do for umoci, and given my experience with documentation not being read by any user ever (in the case of `runc` at least), I think putting text in documentation is mostly a CYA thing than actually having an impact on user behaviour.

estesp:
Just a quick note that we may have a vote close and timezone issue here as @cyphar has made comments and requested changes, but it is now the day of the vote close and I assume @cyphar doesn't stay up all night in his timezone :) So, I'm not sure how that will be resolved prior to noon US Pacific today. I have replaced @vbatts LGTM vote that was lost during the commit squash/edit; I assume no one else has voted yet?

cyphar:
I think it may have been a mistake to run both votes in parallel -- not to mention it's pretty clear we didn't really do a proper review of both proposals until this week (in fact, I have some more questions that I'd like answered about ORAS before I vote).

I think it wouldn't be a bad idea to call another vote on ORAS once we've completed a proper review. I also think that we may have also jumped the gun a bit on the umoci proposal too (given the significant rewrites to clarify many questions I've seen in the past week).

proposals/oras.md
SteveLasker: ORAS was created as a library to support additional artifacts with oci-distribution. Pushing pulling content, creating manifests, .tar-ing up contents isn't trivial for the average person wanting to focus on their unique artifact type and benefit from a registry.

Sure, it happens to implement the spec, but I don't know if I'd call it a reference implementation as mentioned, it does use other libraries as well. We didn't think it needed to duplicate what functionality already existed. 

So, happy to see the text changed to be a library, intended for artifact usage. 
I'd suggest we need libraries with focus on container runtimes, but we can also benefit from libraries that leverage the generic nature of registries to support all types of artifacts.

proposals/oras.md
SteveLasker: Marking experiments that aren't in a versioned/released spec as experimental makes sense. As we enable Notary v2, I expect a bunch of experiments that we'd likely enable.
That said, I don't understand the question on the `config.mediaType`. Can someone elaborate the concern here?

SteveLasker:
I'll propose this: 
When we discussed voting these projects into OCI, the discussion was about the spirit of the projects being applicable to the framing the TOB made for the types of projects that make sense for OCI, vs other bodies like CNCF. 
The edits over the last week are all great edits and feedback. Did we expect the projects to be complete before adopting? 
For myself, I've been really busy this week and didn't expect this much churn to vote the projects on their merits and Josh jumped in quickly, helping as it was his initial PR.
I could see two approaches:
1. Vote the projects based on the framing we made to the TOB, continuing the PRs as any project lives on.
1. Extend the voting another week to address the feedback.

SteveLasker:
LGTM
Based on the spirit of adoption of the project, with additions & edits as with any project.

estesp:
I could be swayed either way (delay vote conclusion by one week **or** try and get people to vote in the next 3+ hours with @SteveLasker's reasonable point about ongoing discussion/PRs to clarify any open questions). I think the issue is at this point I assume the vote may fail given 2 participants will be in the middle of their night at vote close.

I would like everyone on the TOB to feel comfortable that any larger questions are put to rest (with total agreement that there will always be nuance and changes over time that can be handled post-vote), and it feels like trying to get a complete vote will effectively have to be rushed at this point.

It is a bit troublesome that after months of discussion most questions have come up in the 48 hours before the vote, but that's life in open source! :) Agree with @cyphar that given a lot of questions were raised on both proposals running them in parallel may have added to the difficulty there.

Thoughts? Opinions?

SteveLasker:
With no downside to extending, and in the spirit to make everyone comfortable with their questions, _I'd vote to extend the vote_ a week to make it imminent to cause focus, but time to iterate.

proposals/oras.md
SteveLasker: Talking with Sam, the question was related to the implicit use of annotations, not `config.mediatype` and how the use of annotations could imply a standard, instead of something used for the bin prototype.  And, the experimental references which we agree with as well.
We'll discuss removing the annotations, possibly placing in the config object. Tracked at https://github.com/deislabs/oras/issues/167
We'd likely leave the helper capabilities for artifact authors to add annotations for their specific usage, but avoid always generating, implying a standard.

samuelkarp:
I think extending the vote another week to settle on the questions makes sense.

cyphar:
:+1: on extending the vote, and I don't think we need to vote to extend the vote (that'd get a bit silly pretty quickly) -- I think @estesp as TOB Chair can just announce that the vote will be extended since he called it in the first place.

SteveLasker:
I've extended the vote to next Friday. 
We have a meeting scheduled Monday evening to discuss the various issues on annotations and library/reference implementation. 
While I scheduled it for ORAS maintainers, anyone is welcome: 
- Monday 5:00-5:30 Pacific Time
- https://us04web.zoom.us/j/73854402181?pwd=TzE5Vkl6bHhadHk0OC9hL2E1SVdpUT09

estesp:
Thanks @SteveLasker; I did pass the extension by Amy and Chris A. today to make sure there were no governance issues I was unaware of with a decision to extend. I will send a note to the TOB list just to make sure it is clear.

estesp:
What are the remaining questions from TOB members? We need to finalize questions today (and a maximum of maybe 24 hrs from now IMHO) so that answers can be provided and we give the ORAS experts at least a day to provide answers.

I haven't seen any substantial feedback or responses since we extended the vote, but we need a conscious acknowledgement from TOB members that their specific questions are answered and they are all comfortable to make a voting decision.

README.md
```diff
@@ -26,6 +26,7 @@ https://groups.google.com/a/opencontainers.org/forum/#!forum/tob (tob@opencontai
 * [Image Format Spec](proposals/image-format)
 * [SELinux](proposals/selinux.md)
 * [Tools](proposals/tools.md)
+* [ORAS](proposals/oras.md)
```
estesp, line 29: @jdolitsky if you could rebase on master to pick up the conflict in this list (umoci was merged) that would be good so this is mergeable when necessary.

proposals/oras.md
samuelkarp: @cyphar Can you help me understand more about how you see ORAS as an implementation of the distribution-spec?  I can see how ORAS enables people to work more easily with distributions, but ORAS itself (at least in my understanding) doesn't implement the distribution protocol.

proposals/oras.md
SteveLasker: @cyphar It looks like Josh's update addresses the feedback. Is there something else to clear the requested change?

cyphar:
Sorry for not sending this earlier, but I've been struggling to get ORAS to work so I can properly review it. Namely, unless I'm mistaken it appears ORAS cannot actually pull OCI images at all:

```ShellSession
% oras pull docker.io/library/debian:latest -ao oci_store
Downloaded empty artifact
Pulled docker.io/library/debian:latest
Digest: sha256:46d659005ca1151087efa997f1039ae45a7bf7a2cbbe2d17d3dcbda632a3ee9a
% ls oci_store
ls: cannot access 'oci_store': No such file or directory
```

The digest is correct, but I don't know where the blobs are supposed to have gone (I found that ORAS does support using an OCI image-spec layout as of https://github.com/deislabs/oras/pull/108 -- though that really should be the default IMHO). I imagine this can't be intentional, hence why I've put off replying until I got it working (I assumed that I was doing something wrong). But with the vote deadline being so close now, it's probably better to post my incomplete opinion on this proposal.

Regarding "library" or "reference implementation", this project can only be one of the two. A reference implementation may provide a library interface (as `umoci` and `runc` do, to varying degrees) so that's not really the sticking point for something to be a library. Whether something is a library is more about its purpose, size, and other such factors (and I agree with @estesp that opining about that here probably isn't helpful).

However, @SteveLasker has discounted this project being a library (I'm not sure if you saw this @estesp), and instead saying that it is a reference implementation:

> The vote here agrees to merits of ORAS being a reference client implementation of the OCI distribution-spec and OCI Artifacts

Which means it's very critical that I ask whether there are specific plans to make `oras pull docker.io/some/oci-image into/this/oci-image/layout` work out of the box? Given that this is what most people are going to use the distribution-spec for, I think it's fair to say that this should work and the artefacts features (while obviously important) should be something you have to explicitly specify. Namely:

 * The whole `-a` option is a bit concerning (while the artefacts-spec might use layers as a somewhat dubious way to reference other blobs -- the image-spec really should take precedence here when it comes to which mode is easier to use from the CLI). `oras pull` should pull all of the blobs, at least for OCI images. I saw https://github.com/deislabs/oras/issues/130 but that seems to narrowly-scoped -- the issue is more generally that OCI images should be the default (or at least, if there is a default mode it should be OCI images).

 * The storage should be an OCI image-spec layout by default, because that is the primary way that OCI images are stored on local machines. Without this, a user couldn't seamlessly do `oras pull ... ; umoci unpack ... ; runc run ...` which would be the ideal workflow.

I don't think all of the above needs to be addressed before the project is accepted, but there should be broad agreement that this is what ORAS should aim towards in the future. Without the above points settled, users will have to be told "yeah we have a reference implementation of the spec -- ORAS -- but if you actually want to download OCI images use skopeo" which is a very split-brain message.

Again, sorry for springing this on you so late in the process but I do think this is important to clear up.

SteveLasker:
Thanks Aleksa, 
I really appreciate all the detailed feedback. This is the discussion we've been trying to narrow in upon, container runtime tooling, vs. tooling for other artifacts to leverage distribution with Artifacts.

ORAS is NOT intended to be yet another runtime-container tool. It's not focused on the runtime aspects of the image-spec. It's really focused pushing additional content into the distribution-spec, using the manifest format for defining a thing, and the layers that make up that thing. See [Artifacts Scope](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md#scope)

We can certainly figure out how to fix the default OCI Image pull experience. However, the conversation for vote should be focused on:

Enabling new uses of the distribution spec. Enabling the community to build new things they need to distribution.
ORAS enables custom `artifact-cli push/pull` experiences. We do this by providing a set of libraries that implement all the required functionality to enable customers building OCI Artifacts to:

- login to a distribution-spec implementation
- push content to a distribution-spec implementation
- pull content from a distribution-spec implementation

In addition to reference implementation and library, perhaps we're talking about the tool category.

Let me address the specific topics raised for details:

> The digest is correct, but I don't know where the blobs are supposed to have gone

ORAS isn't intended to account for container-runtimes. It's intended to account for file based artifact types, including directories. The recent conversation on annotations will address this. I don't know how runtime images are tar'd up, but if they have a filename or directory, ORAS would support this model. Right now, I'd consider it a non-supported scenario, or a bug we could support if it were felt to be important. Remember, the ORAS binary is a prototyping binary for building your custom `artifact-cli push/pull` experience.

> However, @SteveLasker has discounted this project being a library (I'm not sure if you saw this @estesp), and instead saying that it is a reference implementation:

I'd suggest we need a bit more clarity on what makes a library vs. reference implementation. 

Rather than require an artifact author to start with a go project, adding all the libraries, we provide a _client reference implementation of the distribution-spec_, which supports arbitrary artifact types.

When the artifact author is ready, they can create a go project and include the same libraries to build their `artifact-cli`.

So, this could be a reference implementation that has the libraries to copy/paste/edit your own.
I'm really open to calling it either _library_, _client reference implementation_ or _tool_.

>...ask whether there are specific plans to make `oras pull docker.io/some/oci-image into/this/oci-image/layout` work out of the box? Given that this is what most people are going to use the distribution-spec for, I think it's fair to say that this should work and the artefacts features (while obviously important) should be something you have to explicitly specify.

Noted above, we can enable this, if we feel it's important. While most currently use runtime, ORAS and Artifacts are all about expanding what people are doing tomorrow. ORAS was not intended to be a generic container-runtime tool. We have docker and others for this ORAS focuses on leveraging registries for new artifact types. 

> The whole -a option is a bit concerning...
> I saw deislabs/oras#130 but that seems to narrowly-scoped -- the issue is more generally that OCI images should be the default (or at least, if there is a default mode it should be OCI images).

ORAS is not intended to default to oci-runtime images. See: [https://github.com/deislabs/oras/issues/115](ORAS mediaTypes should default to unknown #115) It's everything else ORAS enables.

Consider ORAS a prototyping tool. When you create `artifact-cli`, you would embed the specific `manifest.config.mediaType` in that cli. The same way `docker build` sets the manifest.config.mediaType to `...vnd.docker...`

Honestly, I'd like to remove the `-a` option. Once you specify an artifact with it's name, it should pull regardless. 

> The storage should be an OCI image-spec layout by default, because that is the primary way that OCI images are stored on local machines. 

> should be broad agreement that this is what ORAS should aim towards in the future. Without the above points settled, users will have to be told "yeah we have a reference implementation of the spec -- ORAS -- but if you actually want to download OCI images use skopeo" which is a very split-brain message.

These points are focusing on the image-runtime being the priority. ORAS is about all the other artifact types.

cyphar:
@SteveLasker 

I think you're conflating the term "container runtime" with quite a few other things, which is understandable given the term's more generic use within the wider industry. However, **within OCI** a container runtime is a very specific thing -- it is a program or service which implements the [OCI Runtime Specification](https://github.com/opencontainers/runtime-spec). Images are not part of the container runtime, so when you said:

> What ORAS isn't:
> *  A container runtime client

I was a little confused, because nobody should expect that to be the case (it's not going to spawn containers based on OCI runtime-spec bundles, so of course it isn't a container runtime). I just want to make sure that we're understanding each others' concerns and using a term like "container runtime" is a bit confusing when it has a specific meaning OCI but isn't being used that way is confusing (at least to me). In particular, this line also seems a bit confusing to me (again, "runtime image" is a term that I think you're using to refer to the OCI image-spec?):

> I don't know how runtime images are tar'd up, but if they have a filename or directory, ORAS would support this model.

Unless I'm mistaken, this is already implemented by deislabs/oras#108. If that mode of operation was the default, or was used based on the media-type (if it matched the OCI image-spec ones) then I would be totally okay with this because it would do what IMHO most users would expect. I think you might be talking about the `org.opencontainers.image.title` annotation here? Each `layer` is a tar archive of the container's root filesystem, they don't have individual file names -- and the usage of `org.opencontainers.image.title` is a little questionable but let's not get into that. I most definitely would **not** expect ORAS to extract the tar archives, but I would expect it to download them into some content-addressible store so that a tool like `umoci` can operate on them -- which is what deislabs/oras#108 appears to do.

But I guess your point is that ORAS has nothing at all to do with OCI images directly, it just so happens that you can in theory (though apparently not practically -- see above) use it with OCI images. That's fine. But distribution-spec *explicitly* does have a lot to do with OCI images. Yes, you can store other artefact types (and that's perfectly fine) but the primary usage people have today (and likely will always have) is related to OCI images. I don't think it's unreasonable to ask "why doesn't that work at all with ORAS right now?".

I'm genuinely not trying to be difficult or anything, I just want to make sure we don't end up in a situation where we have a project in the OCI that explicitly wishes to not primarily operate on OCI images (which is fine), but at the same time is defined as a "reference implementation" of a specification which most people wish to use with OCI images. Because in that situation, if a future TOB had to consider adding a new project that does allow you to push/pull OCI images it would almost certainly be rejected because "hey, we already have ORAS".

I'm not even sure if the problem is related to the "bucket" we put ORAS in -- when someone comes to the OCI and says "oh, I want to use this neat OCI thing -- let me go download an image to go run a container" our answer to that is going to be "well, we have ORAS but it doesn't actually do that -- you need to go elsewhere or write your own tool based on ORAS". That doesn't seem like a reasonable path forward for me.

> While most currently use runtime, ORAS and Artifacts are all about expanding what people are doing tomorrow. ORAS was not intended to be a generic container-runtime tool. We have docker and others for this ORAS focuses on leveraging registries for new artifact types.

But Docker isn't being proposed for inclusion into the OCI. ORAS is. If we just say that ORAS is a "reference implementation of the artefacts spec" then that's one thing, but having a project that looks and smells like a tool that could be used to download OCI images (but can't) without having a project that can seems like a bit of a bad idea -- we'd be ignoring what the vast majority of users actually want to do.

SteveLasker:
Lots of good thoughts, that I didn't read all. It's the end of the day, so for now, I'll say:

- _words are important_
- _words are hard_

I'll review tomorrow in detail and respond. I'm not sure how that impacts the voting timeline, but we're having the right conversation to narrow it in.

estesp:
@cyphar at this hour in my timezone there is way too much text to respond to for completeness, but let me summarize that it seems like there is a huge disconnect when the ORAS proposal goes out of its way to use phrases like "custom content", "artifacts", "other types of content"  and consistently never talks about implementing any functionality related to the **image-spec** in reference to ORAS, yet almost the entire review revolves around (drum-roll please): ORAS doesn't seem to work with the **image-spec**! Yes, exactly. We already have plenty of tools for that all over the container ecosystem.

ORAS is a library to handle the client-side interactions with any given **distribution-spec** implementation, and provides a common library of functionality on which people can implement and explore the ideas and media types being hashed out in the OCI artifacts project.  It has no default purpose, nor would it make sense for it to handle pulling OCI image-spec bundles as it was never designed or even recommended as a tool for that purpose. It would take about 15 minutes of code to use ORAS to make an image-spec "compliant" client (I've actually done this with ORAS and a custom image handler function which properly walks the image-spec), but given that wasn't the purpose, I'm confused as to why that has to be the "default mode" of ORAS when it wasn't designed to be a container image client.

cyphar:
@estesp 

I really should try to make more comments more brief, sorry about that.

I do now better understand that ORAS explicitly does not wish to be related to OCI images. That's fine, but if we're going to throw around the word "reference implementation" in the context of a distribution-spec client I do want to make sure we agree what that actually means. While most of this text pre-dates the artefacts-spec, the distribution-spec explicitly talks about OCI images and their storage. It seems very odd to me that we would say that "a reference implementation of a distribution-spec client" doesn't actually do anything with OCI images.

All I want to avoid is that we end up in a situation where we:

 * Cannot accept a new project into OCI that implements the necessary bits to download an OCI image; and
 * Have a project in the OCI that is millimeters away from doing that, but has defined it out of their scope.

If we can avoid that (even if it's just saying "we don't discount the addition of other projects that implement OCI image-spec specific download features"), then I'd be fine with this proposal. Honestly I'd even be okay with a tiny `oras-image` program that is part of the ORAS repo that implements the ~35 lines needed to be able to download an OCI image. Or `oras oci-image pull` or something.

I'm not sure if I'm missing something, but am I really the only person who finds it strange that you cannot download an *OCI image* from an *OCI distribution-spec repo* with a tool that we are discussing adding to the OCI as a *reference implementation of the OCI distribution-spec*?

> It would take about 15 minutes of code to use ORAS to make an image-spec "compliant" client (I've actually done this with ORAS and a custom image handler function which properly walks the image-spec)

This was my impression as well, looking at the examples and parts of the implementation. That's why I'm finding this all a little bit strange, as though I'm missing some greater context.

SteveLasker:
Thanks @estesp, and @cyphar 
We can add code that downloads an OCI image (image meant to be run as a container through containerd/docker/moby). It's not out of the scope, just not the primary scope.

However, downloading an oci-image is needed for air-gapped environments where someone wants to download content, doesn't have or want the docker client as it requires a daemon, and needs to send the content through some gateway to the air-gapped environment. We're seeing this in IoT scenarios that implement [Purdue networks](https://en.wikipedia.org/wiki/Purdue_Enterprise_Reference_Architecture).

However, oci-image is not the primary scenario and should not be the default for push because:

- We don't want developers who are experimenting pushing content to a registry that looks like a runtime image, `manifest.config.medaiType=application/vnd.oci.image.config.v1+json.` but will fail security scans, deployments and such. It's like pushing a `.zip` file as `.txt` just to get it through email attachment security scans. It's the wrong intent. See: [Defining a Uniquey Artifact Type](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md#defining-a-unique-artifact-type):
  > **Note:** The `config.mediaType` of `application/vnd.oci.image.config.v1+json` is reserved for artifacts intended to be run and instanced by docker, containerd and other OCI Image runtimes and tool chains.

- I believe we should remove the `-a` parameter as a single tag/digest can't actually reference different artifact types: [oras pull shouldn't require --media-type or --allow-all #130](https://github.com/deislabs/oras/issues/130)  

### To summarize:
- We are supportive of adding `application/vnd.oci.image` support. Whether it expands the tar, or lays the tar on disk is a great discussion.
- The default for push would be `application/vnd.unknown.config` per: [ORAS mediaTypes should default to unknown
#115](https://github.com/deislabs/oras/issues/115)
- `-a` should be removed: [oras pull shouldn't require --media-type or --allow-all #130](https://github.com/deislabs/oras/issues/130)
- ORAS is _client_ reference implementation of the distribution-spec, that supports manifests defined in the image-spec, to login/push/pull any type of oci artifact. We can add `application/vnd.oci.image` per above. Users can use it directly, or copy the repo to implement their own `artifact-cli` for login/push/pull, similar to Helm, Singularity, OPA and many others.
- We can decide to categorize ORAS as a _client reference implementation_, library or tool. I'm not overly passionate about the category. My passion and motivation is to make it super easy for the community to continue expanding the ecosystem of things they can safely manage from dev-->production with the long list of security features all registry vendors are implementing.

README.md
jdolitsky: @estesp done

jdolitsky:
I feel this has been filibustered :)

In my opinion, the value of ORAS is not related to images or the image-spec, but rather that it can publish other types of content to registries. It plugs into registries powered by Docker distribution, regardless of OCI specs in play.

To that end, I can see CNCF as an alternative home for ORAS, considering it's built on containerd and used for Helm and Open Policy Agent (all CNCF projects). On the other hand, that might also signal that it has no use outside the "cloud-native" realm. In any case, ORAS can change its course to better represent its employer. It's certainly operating in a gray area right now.

vbatts:
agreed. I tried to follow, but I think once the words began to spill: "when it rains, it pours".

As to CNCF vs OCI, this lends to the funny relationship that exists between these two groups already.
I personally feel that oras makes more sense to live near distribution-spec (despite leaning particularly on code from containerd).

As for pulling OCI images, that should be expected behaviour.

jdolitsky:
Should this be closed? Does opencontainers want oras? Otherwise I offer 3000 dogecoin for it

SteveLasker:
>Otherwise I offer 3000 dogecoin for it
lol, 
We took the feedback to breakup ORAS to be more specific in the contents it consumes. There were a few open issues to cleanup the separation, including: 

- https://github.com/deislabs/oras/issues/181

So, the action item is to move some of the libraries to oci/articacts and streamline the ORAS project. That said, everyone has been pretty busy with Notary v2, Docker TOS and other priorities that we haven't gotten back to this.

My belief is the Notary v2 implementation will require these changes and it will become a forcing function. 
Whether it means we maintain this as open, close it and re-open, or open a new issue, is a good question.

For the sake of cleanliness, I'd be fine with closing this and creating a new issue when we're ready.

jdolitsky:
Hi all - I've added to this week's call agenda the possibility of continuing with this effort. Perhaps we can discuss what requirements people would require to see in oras to make it a suitable opencontainers project? cc @cyphar @samuelkarp

SteveLasker:
We had some action items from the last discussion:
- https://github.com/deislabs/oras/issues/181
- https://github.com/deislabs/oras/issues/178

It would be good to capture the specific steps to come to a consensus on what's required as I believe we agreed it fits the charter of OCI, but needed some cleanup.

PR #76, opened 2020-05-12
Draft of a "getting started" document for the OCI

estesp:
Draft of a document based on our [March and April TOB calls/discussion](https://hackmd.io/kKl1ECKnSLWhgk7dZ2WUFQ#OCI-TOB-23-March-Call)

Please note that I'm not proposing this is the exact spot for this, but I find the GH review cycle maybe more helpful than continuing in HackMD. We can determine appropriate location for this after discussion on the content.

Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com>

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -1,0 +1,127 @@
+# Getting Started
+
+While the [OCI Charter](./CHARTER.md) is our official governing document, many
+people have asked for a gentler introduction to the Open Container Initiative as
+well as general guidance on how to interact with the projects and specifications
+found under the OCI umbrella. This document attempts to be that starting point
+for those who may be new to the OCI, interested in participation, or who want to
+understand if a project is a fit for inclusion or contribution to the OCI.
+
+## What is the OCI?
+
+As our [website](https://opencontainers.org) states, chiefly the OCI is "an open
+governance structure for the express purpose of creating open industry standards
+around container formats and runtime." Created in 2015, the core initial purpose
+was to align Docker and other ecosystem players around a runtime and image
+specification that would become an industry baseline for how images are packaged
+and executed in OCI compliant runtimes.
+
+Since that initial work, finalized in 1.0 specs in the summer of 2017, a
+distribution specification is now underway, based on the HTTP API of the initial
+Docker distribution (registry) project. In 2019, an "artifacts" project was
+approved by the TOB and will be used to define other artifact types (in addition
+to OCI images) which may be defined and stored in registries conforming to the
+OCI distribution spec.
+
+## Who uses the OCI?
+
+It makes the most sense to see consumers of OCI as fitting into a few specific
+categories: contributors/members, implementors, and end users.
+
+**Contributors**, including the project maintainers, and member companies have a
+vested interest in bringing forward the "state of the art" with respect to the
+scope of the OCI. Currently this is of course limited to specifications around
+runtime, image, and distribution of containers. The artifacts repository and
+project is related to both image and distribution and is a natural expansion
+of OCI into ongoing experimentation with registries and cloud native tooling.
+Specifically, [artifacts](https://github.com/opencontainers/artifacts) expands the concept around the **what** when
+discussing non-OCI image content (Helm charts, Singularity container images,
+SPDX, source bundles, etc.) and registries.
+
+**Implementors** own projects outside the OCI for which they have determined
+value in being "OCI compliant"; whether that's a registry, a unique container
+runtime, or an end-user tool that talks to registries. They take the OCI
+specifications and implement support for them in their project, and potentially
+will use the conformance suite within the OCI to validate their compliance to
+the specification(s).
+
+**End Users** tend to gain value from the OCI specifications in an indirect
+way: they will expect projects and products that claim OCI compliance to
+interoperate smoothly with other projects and products which are OCI compliant.
+They will look to the OCI to continue maturing conformance and specifications
+to best support the cloud native ecosystem's goal of interoperability for
+runtimes, images, and distributing images and artifacts across clouds and
+platforms.
+
+## What Types of Projects Exist within the OCI?
+
+There has been some growth in the nature and use of the OCI since those
+initial meetings around the image and runtime specification in 2015. The
+following subsections define project categories which exist in the OCI today.
+
+### Specifications
+
+Clearly the image, runtime, and distribution specifications are the key
+reason for the existence of the OCI today. These standards are meant to
+provide a baseline to follow for implementors of runtimes, registries, and
+tools which interact with container images and registries.
+ - [Runtime spec](https://github.com/opencontainers/runtime-spec)
+ - [Image spec](https://github.com/opencontainers/image-spec)
+ - [Distribution spec](https://github.com/opencontainers/distribution-spec)
+
+### Spec Conformance Test
+
+Conformance tests should provide a clear, end-user runnable test suite for
+implementors to use to determine and demonstrate that an implementing project
+or product meets the explicit definitions of the specification.
+ - [OCI Conformance](https://github.com/opencontainers/oci-conformance)
+
+The most advanced conformance implementation to date is for the new distribution specification. Additional work on image and runtime conformance is ongoing.
+
+### Libraries 
+
+While hosting library code is not a common goal of the OCI, there are a few
+specific cases where small, non-end user focused, and tightly scoped libraries
+have been accepted into the OCI. The common denominator for these libraries are
+that they help implementors properly use features of the specifications:
+ - [go-digest](https://github.com/opencontainers/go-digest)
+ - [selinux](https://github.com/opencontainers/selinux)
+
+Utilities and end-user UX-oriented code is most likely better targeted at other
+more broad communities like the [CNCF](https://cncf.io). While there are not
+explicit rules, a discussion with the TOB is warranted for projects looking to
+contribute a library to OCI.
+
+### Reference Implementations
+
+While theoretically a specification can have one or more reference
+implementations, the OCI runtime spec and the program, `runc`, have gone
+hand in hand simply due to the particulars around the founding of the OCI.
+
+It is not inconceivable that more reference implementations would be 
+contributed and supported within the OCI, but at this point, the only
+active and viable reference implementation within the OCI is the `runc`
+implementation of the runtime specification, based around the contributed
+**libcontainer** codebase contributed by Docker.
+ - [runc](https://github.com/opencontainers/runc)
+
+Runc is also unique in that it is an open source reference implementation,
+but also a core dependent ecosystem component underneath the majority of
+container engine implementations in use today.
+
+For any future reference implementation to be adopted by the OCI, it would
+need to be kept in sync with the specification it implements. For a change
+to be accepted to the spec, the equivalent implementation would need to be
+authored and accepted as well.
+
+## Should my Project be in the OCI?
+
+The OCI receives proposals suggesting additions to the current suite
+of project types listed above. We understand that a perfect framework
+for determining inclusion or rejection of these proposals is an
+intangible goal. However, we list the following considerations to help
+guide future potential submissions.
+
+ 1. The OCI, unlike the CNCF, is not directly chartered for the advancement, marketing, and support of general cloud native software projects.
+ 2. Projects consumed via a UI and/or with a significant end user experience component are unlikely to be a good fit for the OCI model.
+ 3. The OCI has an small but active and vibrant group of participants today; however the specifications and related projects are a small niche of the overall cloud native world, and seeking out the OCI to validate or grow a community around a young project is unlikely to be a viable model. The CNCF is much more suited to that aim given the sandbox and maturity model.


GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+**libcontainer** codebase contributed by Docker.
+ - [runc](https://github.com/opencontainers/runc)
+
+Runc is also unique in that it is an open source reference implementation,
+but also a core dependent ecosystem component underneath the majority of
+container engine implementations in use today.
+
+For any future reference implementation to be adopted by the OCI, it would
+need to be kept in sync with the specification it implements. For a change
+to be accepted to the spec, the equivalent implementation would need to be
```
samuelkarp, line 114: I think it'd be worth articulating the converse here as well; for a change to be accepted to the reference implementation, the equivalent change would need to be adopted in the spec as well.  Runc hasn't followed this, and that means that integrators with runc may be integrating with its specific behavior rather than with the behavior defined by the spec.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+
+It makes the most sense to see consumers of OCI as fitting into a few specific
+categories: contributors/members, implementors, and end users.
+
+**Contributors**, including the project maintainers, and member companies have a
+vested interest in bringing forward the "state of the art" with respect to the
+scope of the OCI. Currently this is of course limited to specifications around
+runtime, image, and distribution of containers. The artifacts repository and
+project is related to both image and distribution and is a natural expansion
+of OCI into ongoing experimentation with registries and cloud native tooling.
```
SteveLasker, line 36: The word experimentation is creating angst amongst some that see this as experimentation of registries that gives them the belief this will change. 
What I think you're trying to say is supporting other types, not as an experiment, but a commitment to support other types, enables others to experiment on a stable capability.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+## Should my Project be in the OCI?
+
+The OCI receives proposals suggesting additions to the current suite
+of project types listed above. We understand that a perfect framework
+for determining inclusion or rejection of these proposals is an
+intangible goal. However, we list the following considerations to help
+guide future potential submissions.
+
+ 1. The OCI, unlike the CNCF, is not directly chartered for the advancement, marketing, and support of general cloud native software projects.
+ 2. Projects consumed via a UI and/or with a significant end user experience component are unlikely to be a good fit for the OCI model.
```
SteveLasker, line 126: 1 really helps scope what you get with OCI. 
I'm struggling with 2. I think I see what you're saying isn't a good solution. Could we focus on what would? Supportive libraries or tools that enable the OCI specs, that need some level of endorsement to gain acceptance.

GETTING_STARTED.md
cyphar: Maybe, but there is a necessary balance here.

It would be a mistake to accept changes into a specification if they haven't been tested "in the wild" (or at the very least, implemented at least once). The image-spec had this exact problem -- no reference implementation (at least, until umoci) made it so that we had to do some pretty large reworks during the release-candidate phase of `1.0`.

That being said, I think that we made too much of a concession to runc in the runtime-spec (though I would argue this is also historically driven -- runc was ready and in wide use quite a while before the spec was). There are many things in runc which really should be in the spec, and I am constantly frustrated to see that many implementations of the runtime-spec spend more of their time reading the runc source code than they do the spec.

Maybe we should make it so that no *release* of the project should have any major not-in-the-spec features, which would allow for experimentation and development without risking people becoming dependent on the feature. Unfortunately runc's behaviour doesn't really break this rule (because it breaks the entire concept of releases given that we've been doing `1.0` release candidates for 4 years now) -- but I think it's fair to say that this is something which no other OCI project will or should repeat.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+ 3. The OCI has an small but active and vibrant group of participants today; however the specifications and related projects are a small niche of the overall cloud native world, and seeking out the OCI to validate or grow a community around a young project is unlikely to be a viable model. The CNCF is much more suited to that aim given the sandbox and maturity model.
```
cyphar, line 127: I think this should be merged into (2) as being part of the "boring infrastructure" point. In fact, I think most of the guidelines could be considered to be part of the "boring infrastructure" point -- maybe we should break it down like this:

> The OCI receives proposals suggesting additions to the current suite of project types listed above. We understand that a perfect framework for determining inclusion or rejection of these proposals is an intangible goal.  However, we list the following considerations to help guide the review of future project submissions.
>
> * The project should be a piece of "boring core container infrastructure". This is a partially subjective criterion, but the key factors that make something fulfill this requirement are:
>
>   1. The project should be as un-opinionated and extensible as is reasonably possible.
>   2. Rather than being a complete solution or framework, the project should be usable as a building-block for larger solutions and frameworks.
>   3. (probably more things go here...)
>
> * It should fit logically into the scope and mission of the OCI -- to provide a home for open standards and tools which underpin the wider container ecosystem. This means that it should not conflict with existing OCI projects, nor should be it be completely unrelated to existing OCI projects. The precise scope of the OCI is defined by the OCI Charter, but the TOB may choose to expand the scope if they feel it is within the mission of the OCI.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Should my Project be in the OCI?
+
+The OCI receives proposals suggesting additions to the current suite
+of project types listed above. We understand that a perfect framework
+for determining inclusion or rejection of these proposals is an
+intangible goal. However, we list the following considerations to help
+guide future potential submissions.
+
+ 1. The OCI, unlike the CNCF, is not directly chartered for the advancement, marketing, and support of general cloud native software projects.
```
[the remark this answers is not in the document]
AkihiroSuda: What does "support" mean?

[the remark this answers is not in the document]
cyphar: I presume it means enterprise-like support, but the CNCF doesn't provide that either...? @estesp?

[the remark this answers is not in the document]
estesp: I specifically was thinking of the project support staff provided by the CNCF along with a special service desk which provides help with website development, cloud/infrastructure cost coverage/free credits, social media reports for member projects, etc. etc. I think the careful line here is that the OCI has a budget for some of these items as well, but the point is that it isn't as clearly in line with the mission of OCI to gain a large # of projects and then provide that support for their maintainers/communities as it seems to be specifically in line with the CNCF mission.

estesp:
Changelog for 4 June 2020:
 - added note specifying precedence of charter over this "informal" document
 - reworded first para of "What is the OCI?" to allow for mission statement to be different today
 - clarified artifacts as not an "experimental" thing, but a stable base on which experimental media types can be validated/tested/etc.
 - added artifacts to "specifications" section with language around how we might see it there
 - used @cyphar's content directly for the "Reference Implementations" section as I thought it was a solid improvement
 - Lots of reword to handle comments/suggestions around "Does my project fit?" section; used some of @cyphar's suggested commentary there as well.

Thanks to @SteveLasker, @AkihiroSuda, @cyphar, and @samuelkarp for comments/review!

GETTING_STARTED.md
estesp: Given this section was updated due to further review below from @cyphar, there is no more commentary on spec vs. ref. implementation alignment. Not that it isn't an important topic in general, but maybe it didn't fit in this document anyway? Feedback welcome.

GETTING_STARTED.md
estesp: Updated in latest commit.

GETTING_STARTED.md
estesp: This section is updated significantly; would love your re-review/comments.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
+# Getting Started
+
+While the [OCI Charter](./CHARTER.md) is our official governing document, many
+people have asked for a gentler introduction to the Open Container Initiative as
```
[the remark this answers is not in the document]
estesp: Good question; let's discuss on the TOB call.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+well as general guidance on how to interact with the projects and specifications
+found under the OCI umbrella. This document attempts to be that starting point
+for those who may be new to the OCI, interested in participation, or who want to
+understand if a project is a fit for inclusion or contribution to the OCI.
+
+## What is the OCI?
+
+As our [website](https://opencontainers.org) states, chiefly the OCI is "an open
+governance structure for the express purpose of creating open industry standards
+around container formats and runtime." Created in 2015, the core initial purpose
```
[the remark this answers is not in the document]
estesp: Reworded to make this paragraph hopefully just a brief history statement and not tied to whatever we might change the mission statement to today or someone else does in the future. Let me know if it reads that way to you.

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+It is not inconceivable that more reference implementations would be 
+contributed and supported within the OCI, but at this point, the only
+active and viable reference implementation within the OCI is the `runc`
+implementation of the runtime specification, based around the contributed
+**libcontainer** codebase contributed by Docker.
+ - [runc](https://github.com/opencontainers/runc)
+
+Runc is also unique in that it is an open source reference implementation,
+but also a core dependent ecosystem component underneath the majority of
+container engine implementations in use today.
```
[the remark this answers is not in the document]
estesp: I like your text and have adopted it for this section in the updated commit.

GETTING_STARTED.md
estesp: Reworded this section and used some of your suggested text above.

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+OCI distribution spec.
+
+## Who uses the OCI?
+
+It makes the most sense to see consumers of OCI as fitting into a few specific
+categories: contributors/members, implementors, and end users.
+
+**Contributors**, including the project maintainers, and member companies have a
+vested interest in bringing forward the "state of the art" with respect to the
+scope of the OCI. Currently this is of course limited to specifications around
```
samuelkarp, line 38: nit: remove "of course"

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+runtime, image, and distribution of containers. The artifacts repository and
+project is related to both image and distribution and is a natural expansion
+of OCI into ongoing support of additional types within registries and related
```
samuelkarp, line 41:
```suggestion
runtime, image, and distribution of containers. The artifacts repository and
project expands the scope of OCI to support different types of content in
concert with the image and distribution specifications.
```

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+cloud native tools. Specifically, [artifacts](https://github.com/opencontainers/artifacts)
+expands the concept around the **what** when discussing non-OCI image content
+(Helm charts, Singularity container images, SPDX, source bundles, etc.) and
+registries, and allows for further experimentation with arbitrary file types
+within the context of this stable andd well-defined capability.
```
samuelkarp, line 46:
```suggestion
(Helm charts, Singularity container images, SPDX, source bundles, etc.) in
registries, and allows for further experimentation with arbitrary file types
within the context of this stable andd well-defined capability.
```

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+runtime, or an end-user tool that talks to registries. They take the OCI
+specifications and implement support for them in their project, and potentially
+will use the conformance suite within the OCI to validate their compliance to
+the specification(s).
+
+**End Users** tend to gain value from the OCI specifications in an indirect
+way: they will expect projects and products that claim OCI compliance to
+interoperate smoothly with other projects and products which are OCI compliant.
+They will look to the OCI to continue maturing conformance and specifications
+to best support the cloud native ecosystem's goal of interoperability for
```
samuelkarp, line 59: nit: I'd avoid "cloud native" as I think it's important for us to consider OCI in the context of non-cloud use-cases.  

```suggestion
to best support the container ecosystem's goal of interoperability for
```

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+ * Proposed projects should fit logically into the scope and mission of the
+ OCI: a home for open standards and tools which underpin the wider container
+ ecosystem. Proposed additions to the OCI should not conflict with existing OCI
+ projects, nor should they be completely unrelated to existing OCI projects.
+ The precise scope of the OCI is defined by the OCI Charter, but the TOB may
+ choose to expand the scope if they feel it is within the mission of the OCI.
+
+In summary, the OCI has a small but active and vibrant group of participants
+today. However the OCI specifications and related OCI projects are but a small
+niche of the overall cloud native world, and seeking out the OCI to validate
```
samuelkarp, line 165:
```suggestion
niche of the overall container world, and seeking out the OCI to validate
```

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+
+## What Types of Projects Exist within the OCI?
+
+There has been some growth in the nature and use of the OCI since those
+initial meetings around the image and runtime specification in 2015. The
+following subsections define project categories which exist in the OCI today.
+
+### Specifications
+
+Clearly the image, runtime, and distribution specifications are the key
```
samuelkarp, line 71: nit: remove "Clearly"; it doesn't really add anything to the sentence

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+ - [Runtime spec](https://github.com/opencontainers/runtime-spec)
+ - [Image spec](https://github.com/opencontainers/image-spec)
+ - [Distribution spec](https://github.com/opencontainers/distribution-spec)
+
+Artifacts is not a standalone specification, but fits within this category
+as it is a repository which serves as the extension point for media types
+which include the core media types found within the image spec, as well as
+connecting those and additional media types with the distribution spec as
+the means for storage, query, and delivery of a growing list of well-defined
+artifact types defined within the project.
```
samuelkarp, line 84: I'm having a really hard time parsing this.  Would something like this make sense?

> Artifacts is not a standalone specification, but provides additional guidance on top of the image and distribution specifications for formatting and distributing content.

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+intangible goal. However, the following considerations are provided
+to help guide potential submissions.
+
+ * The OCI is not directly chartered for the advancement, marketing, and
+ support of general cloud native software projects. As our charter states: *The
+ Open Container Initiative does not seek to be a marketing organization, define
+ a full stack or solution requirements, and will strive to avoid standardizing
+ technical areas undergoing innovation and debate.*
+ * The project should be a piece of "**boring core container infrastructure**".
+ While this is partially subjective criterion, the key factors towards
```
samuelkarp, line 147:
```suggestion
 While this is a subjective criterion, the key factors towards
```

GETTING_STARTED.md
```diff
@@ -0,0 +1,127 @@
...
+specification that would become an industry baseline for how images are packaged
+and executed in OCI compliant runtimes.
+
+Since that initial work, finalized in 1.0 specs in the summer of 2017, a
+distribution specification is now underway, based on the HTTP API of the initial
+Docker distribution (registry) project. In 2019, an "artifacts" project was
+approved by the TOB and will be used to define other artifact types (in addition
+to OCI images) which may be defined and stored in registries conforming to the
+OCI distribution spec.
+
```
SteveLasker, line 30: > In 2019, an "artifacts" project was approved by the TOB and will be used to define other artifact types (in addition to OCI images) which may be defined and stored in registries conforming to the OCI distribution spec.

suggest:
In 2019, an "artifacts" project was approved by the TOB enabling other artifact types (in addition to OCI images). Supporting additional artifacts acknowledges containers are core to modern workflows, while enabling additional artifact types to be stored in registries conforming to the OCI distribution spec

GETTING_STARTED.md
SteveLasker: Not sure cloud native is really tied to the cloud, per se. Yes, literally the word is there. But, is someone using CNCF projects on-prem not "cloud native"? :) This is also the tension we're seeing around OCI being tightly coupled to containers, while the distribution-spec and artifacts enable us to store additional non-container artifacts in a cloud-native/container centric deployment model. 
Anyway, just suggesting this wording cloud native helps our broader cause here.

GETTING_STARTED.md
```diff
@@ -0,0 +1,169 @@
...
+implementors to use to determine and demonstrate that an implementing project
+or product meets the explicit definitions of the specification.
+ - [OCI Conformance](https://github.com/opencontainers/oci-conformance)
+
+The most advanced conformance implementation to date is for the new distribution specification. Additional work on image and runtime conformance is ongoing.
+
+### Libraries 
+
+While hosting library code is not a common goal of the OCI, there are a few
+specific cases where small, non-end user focused, and tightly scoped libraries
```
dmcgowan, line 99: Defining "small" here is kind of difficult. I would like to see language here specifying that these libraries should have no or minimal dependencies. If a library is depending on an even larger project or library, it can't really be "small". I think the two libraries below are good examples of "no or minimal dependencies".

cyphar: I somewhat agree with putting more text about what "small" means, but I don't think this needs to be a legal document in terms of specificity. After all, the TOB decides whether a project meets the criteria we outline here and in the charter text.

I would also point out that in my view, what defines a library as small is its scope (though that is covered by "tightly scoped" so maybe this "small" does refer to size). As an obvious example, glibc doesn't have any dependencies, but it isn't a small library.

PR #81, opened 2020-06-05
CHARTER: word-wrap at 80 characters

cyphar:
This doesn't contain any text changes, it only changes the layout so
that you can effectively read the charter in a terminal. I know it is
"bad practice" in the OCI to have hard line-wrapping (the argument being
that it is harder to review changes), but I've never found that to be an
issue.

### DIFF
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -8,201 +8,454 @@
 
 ##  1. Mission of the Open Container Initiative (“OCI”).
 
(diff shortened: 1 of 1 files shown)


cyphar:
I don't believe this change requires a vote of any kind (aside from the standard two-LGTM review) since the text is identical between the two versions, it's just word-wrapped (and Section 10's header is made the same size as all the others'). I've split this out from my Charter rework so I don't have to keep rebasing it and so that the changes in the eventual rework aren't obscured.

cyphar:
As this isn't a text change I think it's okay to merge with 2 LGTMs -- @caniszczyk?

caniszczyk:
That works for me, you crazy 80 character line wrap folks ;)

PR #97, opened 2021-04-27
CHARTER: s12(c): formalise charter versions and changelog

cyphar:
We have version numbers (and a changelog) in the Charter, so we should
probably actually enshrine that in the Charter. In addition, clean up
the changelog format so that it's easier to understand.

This change does introduce new requirements for the editors of the
Charter, but it simply mirrors the current convention applied to Charter
amendments.

In addition, define the concept of a "draft Charter version" so that we
can make iterative changes to the charter through separate PRs without
having to enact each minor change separately.

Split off from #86.

### DIFF
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -1,12 +1,18 @@
-# Open Container Initiative Charter v1.2
+# Open Container Initiative Charter v1.2+DRAFT
 
-| Comment          | Effective Date   |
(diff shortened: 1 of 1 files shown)


SteveLasker:
Ok, so much for accidental clicking. Where's the undo?

SteveLasker (approved):
LGTM

PR #158, opened 2026-03-23
Define procedures for TOB vacancy appointments

jorydotcom:
This copy addresses #155 / how to handle vacancies.

### DIFF
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -250,6 +250,11 @@ e. The TOB shall be composed of nine (9) individuals elected for their
    entity, or owned by or under common ownership with each other. TOB members
    may not designate alternative representatives.
 
+   In the event of a TOB vacancy, the remaining TOB members may either:
+   - i. Appoint an individual to fill the vacant seat. Appointed individuals serve until the next TOB election, no more than one calendar year from appointment.
+   - ii. Run a new election to fill the vacant seat. Elected individuals serve for the duration of the group term (f), no more than two calendar years from election.
+   - iii. Take no action.
+
 f. TOB members shall be split into two (2) groups, serving for a term of two
    (2) years on a staggered basis, where one group is elected each year. The
    initial TOB will have four (4) TOB members who will only serve for a term of


CHARTER.md
```diff
@@ -250,6 +250,11 @@ e. The TOB shall be composed of nine (9) individuals elected for their
    entity, or owned by or under common ownership with each other. TOB members
    may not designate alternative representatives.
 
+   In the event of a TOB vacancy, the remaining TOB members may either:
+   - i. Appoint an individual to fill the vacant seat. Appointed individuals serve until the next TOB election, no more than one calendar year from appointment.
+   - ii. Run a new election to fill the vacant seat. Elected individuals serve for the duration of the group term (f), no more than two calendar years from election.
```
samuelkarp, line 255:
```suggestion
   - ii. Run a new election to fill the vacant seat. Elected individuals serve for the duration of the group term (f), no more than two calendar years from election. A new election under this option is considered the next TOB election for the purposes of Section 6(e)(i) above.
```

CHARTER.md
```diff
@@ -250,6 +250,11 @@ e. The TOB shall be composed of nine (9) individuals elected for their
    entity, or owned by or under common ownership with each other. TOB members
    may not designate alternative representatives.
 
+   In the event of a TOB vacancy, the remaining TOB members may either:
+   - i. Appoint an individual to fill the vacant seat. Appointed individuals serve until the next TOB election, no more than one calendar year from appointment.
```
tianon, line 254: Is this a 2/3 vote?  Unanimous?  Majority?

(my personal opinion would be that this should be a high bar given it circumvents the normal flow - either 2/3 or unanimous with a slight preference for the latter)

This being one-year means we also have an automatic vacancy at the next election, right?  That wouldn't be filled from that election, because it would be part of the other group term.  That's sane (it lets the new crop participate in the appointment of that empty seat fresh again), but I want to make sure I'm understanding the timeline here.

CHARTER.md
sudo-bmitch: I was confused by this suggestion @samuelkarp. I think it's trying to say "if the TOB appoints an individual, and the original term remaining was for more than a year, then we must run a new election for the seat when there's 1 year remaining in the original term". If we leave this language out, we would still have the other options, giving us more flexibility (perhaps there's a reason we didn't run a new election before).

CHARTER.md
```diff
@@ -250,6 +250,11 @@ e. The TOB shall be composed of nine (9) individuals elected for their
    entity, or owned by or under common ownership with each other. TOB members
    may not designate alternative representatives.
 
+   In the event of a TOB vacancy, the remaining TOB members may either:
```
sudo-bmitch, line 253: We should clarify how the TOB decides which of the three options to use. Perhaps a simple majority vote, that defaults to (iii) no action when a majority cannot be reached?

CHARTER.md
samuelkarp: > I think it's trying to say "if the TOB appoints an individual, and the original term remaining was for more than a year, then we must run a new election for the seat when there's 1 year remaining in the original term"

No, that's not what I was trying to say.

The goal of this language was to allow the TOB to appoint someone immediately _and_ to subsequently run an election that would replace the appointed seat. Section 6(e)(i) says "until the next TOB election, no more than one calendar year" and the clarification here says "new elections even if they are not regularly scheduled still count". This was what @mikebrow had suggested on the call and I wanted to make sure it was captured in the text.

CHARTER.md
sudo-bmitch, line 254:
```suggestion
   - i. Appoint an individual to fill the vacant seat by vote of 2/3rds of the elected TOB members. Appointed individuals serve until the next TOB election, no more than one calendar year from appointment, at which time a seat vacancy is resolved by the next elected TOB. An appointed seat may be filled by a separate election according to Section 6(e)(ii).
```

CHARTER.md
sudo-bmitch: Got it. I added a suggestion for (i) above that feels like an easier read for me. Particularly since it's not clear we can do multiple options with the "either" phrasing on line 253.

CHARTER.md
samuelkarp: I don't think we need to define something separate for this from how TOB makes other decisions (which is already covered in the charter, see Section 6(n)).

CHARTER.md
samuelkarp: Section 6(n) already covers "consensus by default with 2/3 vote threshold"

samuelkarp: Section 6(n) already covers "consensus by default with a 2/3 vote threshold"

sudo-bmitch:
```suggestion
   - i. Appoint an individual to fill the vacant seat. Appointed individuals serve until the next TOB election, no more than one calendar year from appointment, at which time a seat vacancy is resolved by the next elected TOB. An appointed seat may be filled by a separate election according to Section 6(e)(ii).
```

mikebrow:
```suggestion
   - i. Appoint an individual to fill the vacant seat. Unless replaced by election according to Section 6(e)(ii), Appointed individuals serve until the next TOB election, no more than one calendar year from appointment.
```

and remove "either" 
In the event of a TOB vacancy, the remaining TOB members may:

